PR #12, opened 2020-02-07
Authentication pages refactor and redesign

dinohamzic:


app/views/spree/components/buttons/_primary.html.erb
```diff
@@ -1,13 +1,15 @@
...
 
   # Optional props
   disabled = false if local_assigns[:disabled].nil?
+  full_width = false if local_assigns[:full_width].nil?
   name = nil if local_assigns[:name].nil?
   type = "button" if local_assigns[:type].nil?
 
   # Classes
   class_names = [base_class]
+  if full_width then class_names.push("#{base_class}--full-width") end
```
kennyadsl: This can be written as:

```ruby
class_names.push("#{base_class}--full-width") if full_width
```

app/views/spree/components/auth/_change_form.html.erb
```diff
@@ -0,0 +1,40 @@
+<%
+  html = { class: "auth-form" }
+  url = spree.update_password_path
+%>
+
+<%= form_for @spree_user, as: :spree_user, url: url, method: :put, html: html do |f| %>
+  <div class="auth-form__input-wrapper">
```
kennyadsl: Why is this wrapper needed?

app/views/spree/components/buttons/_primary.html.erb
```diff
@@ -1,13 +1,15 @@
 <%
-  base_class = "button".freeze
+  base_class = "button-primary".freeze
 
   # Optional props
   disabled = false if local_assigns[:disabled].nil?
+  full_width = false if local_assigns[:full_width].nil?
```
kennyadsl: I think we should use:

```ruby
full_width = false if local_assigns[:full_width].blank?
```

because as is, it returns `true` even when a `full_width: false` is passed to the partial.

app/views/spree/user_registrations/new.html.erb
```diff
@@ -0,0 +1,9 @@
+<% @body_id = 'signup' %>
```
kennyadsl: do we need to keep these `@body_id` for any reason?

app/views/spree/components/auth/_change_form.html.erb
dinohamzic: To decouple auth form input positioning from the input component itself.

app/views/spree/components/buttons/_primary.html.erb
dinohamzic: I'm not sure I follow. Isn't the logic here: 
- if a value is not provided, use `false`
- if a value is provided, use the value itself

so that if we pass `false`, `full_width` will still be false.

app/views/spree/components/buttons/_primary.html.erb
dinohamzic: Way better, thanks!

app/views/spree/user_registrations/new.html.erb
dinohamzic: Probably not, I'll see what happens if we remove it.

app/views/spree/components/buttons/_primary.html.erb
kennyadsl: if we pass `false`, with current code the line is not executed at full_width is not set. To accomplish that logic you have to use [`.presence`](https://apidock.com/rails/Object/presence)

PR #17, opened 2020-02-21
Product page refactor

dinohamzic:
#21 

This PR is a refactor of the product page (ex: */products/solidus-long-sleeve*) and all its related components, with bare bones styling.

| No variants | With variants |
| --- | --- |
| Screen Shot 2020-03-06 at 12 02 19 | Screen Shot 2020-03-06 at 12 02 34 |


app/views/spree/products/show.html.erb
```diff
@@ -1,35 +1,18 @@
...
-
-    <footer>
-      <%= render partial: 'taxons' %>
-    </footer>
+    <div class="product-page__info">
+      <%= render partial: 'spree/components/products/product_info' %>
+      <%= render partial: 'spree/components/products/product_properties' %>
+      <%= render partial: 'spree/components/products/product_promotions' %>
+      <%= render partial: 'spree/components/products/product_form' %>
+      <%= render partial: 'spree/components/products/product_taxons' %>
```
kennyadsl: We can remove `partial: ` here as well. 

Also, I'd name this files without the `product_` part, since they already are into the `components/products/` folder, and that piece is not really adding anything. Or am I missing something?

app/views/spree/components/products/_product_form.html.erb
```diff
@@ -0,0 +1,63 @@
...
+          <%= display_price(@product) %>
+        </span>
+        <span
+          itemprop="priceCurrency"
+          content="<%= current_pricing_options.currency %>"
+        ></span>
+      </h2>
+
+      <% if @product.master.can_supply? %>
+        <link itemprop="availability" href="http://schema.org/InStock" />
```
kennyadsl: What about moving this availability content into a separate component?

app/views/spree/components/products/_product_form.html.erb
```diff
@@ -0,0 +1,63 @@
...
+  }
+%>
+
+<%= form_for :order, url: populate_orders_path, html: html do |f| %>
+  <%= render partial: 'spree/components/products/product_variants' %>
+
+  <% if @product.price_for(current_pricing_options) and !@product.price.nil? %>
+    <div class="product-submit" id="product-price">
+      <h2 class="product-submit__title">
+        <span
```
kennyadsl: What about moving the price markup into a specific component?

app/views/spree/products/show.html.erb
dinohamzic: The reasoning is that it is way easier to find (in the filesystem and in the code) and reason about, since the aim is to make the template file, scss file and class name all share the same name. What do you think?

app/views/spree/components/products/_product_form.html.erb
dinohamzic: Yes, way better!

app/views/spree/components/products/_product_form.html.erb
dinohamzic: This block of code is used only in this component, maybe in this case it's a bit overkill, as it is also strictly related to the component itself.

app/views/spree/products/show.html.erb
kennyadsl: So the concept is that the file `product_taxons` contains a component with class `product-taxons`, right? If yes, I think it's a good reason to keep it as you proposed.

app/views/spree/components/products/_product_form.html.erb
kennyadsl: I'm ok with this but I also think that it's in this component for layout reasons at the moment. You can theoretically move it anywhere on the page, isn't it? Or use it into the products listing as well. 

For users that want to make things like that it will be much easier to just use the component when they want if it's a separate file.

Anyway, I don't want this to be a blocker, we can even extract it later if we think it makes more sense. Up to you!

app/views/spree/products/show.html.erb
dinohamzic: Correct! Way easier to search.

Issue #18, opened 2020-02-27
[Specs] Convert controller specs into requests specs


kennyadsl:
Using controller specs is now discouraged by both Rails and RSpec teams: https://github.com/rspec/rspec-rails#request-specs, and we already did something similar for Solidus API with https://github.com/solidusio/solidus/pull/2052

PR #22, opened 2020-03-06
System specs

### LINKED ISSUE #19: [Specs] Convert feature specs into system specs
The RSpec team now officially recommends system specs: https://github.com/rspec/rspec-rails#system-specs.

We should make a serious attempt to convert the existing feature specs into system specs.

blocknotes:
Switch from feature specs to system specs.

Details here:
https://relishapp.com/rspec/rspec-rails/docs/system-specs/system-spec

Fixes #19


spec/spec_helper.rb
```diff
@@ -24,6 +24,8 @@
   config.infer_spec_type_from_file_location!
   config.use_transactional_fixtures = false
 
+  config.extend Spree::TestingSupport::AuthorizationHelpers::Controller, type: :system
```
kennyadsl: This should probably go into `solidus_dev_support` right?

spec/spec_helper.rb
```diff
@@ -34,6 +36,14 @@
...
 
+  config.before(:each, type: :system) do
+    driven_by :rack_test
+  end
+
+  config.before(:each, type: :system, js: true) do
+    driven_by :apparition
+  end
+
   config.after(:each, type: :feature) do |example|
```
kennyadsl: This should be switched to `type: :system`. Also, we probably have other similar things happening for `:feature` test only in `solidus_dev_support`. Maybe it's better to change that gem as first step?

solidus_starter_frontend.gemspec
```diff
@@ -33,4 +33,5 @@ Gem::Specification.new do |s|
   s.add_development_dependency 'rubocop-rails'
   s.add_development_dependency 'rubocop-rspec'
   s.add_development_dependency 'rubocop-i18n'
+  s.add_development_dependency 'selenium-webdriver'
```
kennyadsl: Shouldn't this be already added by `solidus_dev_support`?

spec/spec_helper.rb
blocknotes: It could be a good idea, I'm going to ask if I can open this kind of PR there.

spec/spec_helper.rb
blocknotes: I forgot that part. Yes, I need to update it.
But about switching to system specs in dev_support... I'm not convinced.
Some people could think to do this change at another time (or to skip it completely).

solidus_starter_frontend.gemspec
blocknotes: I'm checking, but it doesn't seem to be included there.

spec/spec_helper.rb
kennyadsl: Sorry, I didn't mean switching but just adding support for both system and feature test in `solidus_dev_support`. Using one or another will be a plugin choice

blocknotes: Proposed a change here: https://github.com/solidusio/solidus_dev_support/pull/92

spec/spec_helper.rb
```diff
@@ -34,7 +36,15 @@
...
-  config.after(:each, type: :feature) do |example|
+  config.before(:each, type: :system) do
+    driven_by :rack_test
+  end
+
+  config.before(:each, type: :system, js: true) do
+    driven_by :apparition
+  end
+
+  config.after(:each, type: :system) do |example|
```
kennyadsl, line 47: Shouldn't we conditionally load this only on older Solidus versions after https://github.com/solidusio/solidus/pull/3552 otherwise it's loaded twice when testing against Solidus master?

blocknotes: Good for me 😄 
I added a new commit for this

spec/spec_helper.rb
```diff
@@ -24,8 +24,20 @@
...
+    config.extend Spree::TestingSupport::AuthorizationHelpers::Request, type: :system
+
+    config.after(:each, type: :system) do |example|
+      missing_translations = page.body.scan(/translation missing: #{I18n.locale}\.(.*?)[\s<\"&]/)
+      if missing_translations.any?
+        puts "Found missing translations: #{missing_translations.inspect}\n"
+        puts "In spec: #{example.location}"
+      end
+    end
+  end
```
kennyadsl, line 37: Shouldn't we also do:

```
else
  require 'spree/testing_support/translations'
end
```

? Otherwise in Solidus >= 2.11 (and Solidus master) it won't be loaded, isn't it?

kennyadsl: Wait, I think we need it, but not here in `Rspec.config`. Maybe we should just load the file conditionally at line `#18`?

`require 'spree/testing_support/translations' unless Spree.solidus_gem_version < Gem::Version.new('2.11')`

blocknotes: Good for me, update in the last fixup commit

blocknotes:
Rebased, added a new commit to permit to check the Capybara driver via env variables, made some more checks.

Ready to merge for me 👍

PR #24, opened 2020-03-13
Cart page refactor

dinohamzic:
## Description

This is a refactor and restyle of the `/cart` page and all its components.

## Motivation and Context

This is part of the ongoing refactor effort of the starter Solidus store.

## How Has This Been Tested?

We kept the old Solidus starter store tests and based the refactor on them.

Screen Shot 2020-03-20 at 09 53 34

## Types of changes

## Checklist:

### DIFF
--- a/app/views/spree/components/cart/_cart_items.html.erb
+++ b/app/views/spree/components/cart/_cart_items.html.erb
@@ -1,0 +1,25 @@
+<%
+  order = order_form.object
+
+  is_order_empty = (
+    order.adjustments.nonzero.exists? ||
+    order.line_item_adjustments.nonzero.exists? ||
+    order.shipment_adjustments.nonzero.exists? ||
+    order.shipments.any?
+  )
+%>
+
+<section class="cart-items">
+  <%= render(
+    partial: 'spree/components/cart/cart_item',
+    collection: order.line_items,
+    as: :line_item,
+    locals: { order_form: order_form }
+  ) %>
+
+  <% if is_order_empty %>
+    <p><%= t('spree.cart_subtotal', count: order.line_items.sum(:quantity)) %></p>
+    <p><%= order.display_item_total %></p>
+    <%= render 'spree/components/cart/cart_adjustments' %>
+  <% end %>
+</section>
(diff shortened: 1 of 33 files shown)


app/views/spree/components/cart/_cart_items.html.erb
```diff
@@ -0,0 +1,21 @@
...
+    collection: order.line_items,
+    as: :line_item,
+    locals: { order_form: order_form }
+  ) %>
+
+  <% if (
+    order.adjustments.nonzero.exists? ||
+    order.line_item_adjustments.nonzero.exists? ||
+    order.shipment_adjustments.nonzero.exists? ||
+    order.shipments.any?
```
kennyadsl, line 15: I see this logic eligible for a partial's variable, like `is_order_empty`, WDYT?

app/views/spree/components/forms/inputs/_text.html.erb
```diff
@@ -7,11 +7,13 @@
   label = false if local_assigns[:label].nil?
   min = false if local_assigns[:min].nil?
   placeholder = "" if local_assigns[:placeholder].nil?
+  small = false if local_assigns[:small].nil?
```
kennyadsl, line 10: I'm not really onboard with this: being small is something that works for our design but someone else could want to have that thing (also applies to button component) bigger. What about moving these rules to CSS only?

app/views/spree/components/cart/_cart_items.html.erb
dinohamzic: Great idea, much cleaner 👍

app/views/spree/components/forms/inputs/_text.html.erb
dinohamzic: I'm not sure I'm following. `small` is toggle-able. If the user wants the button to be bigger (normal/default size), just don't pass anything when rendering the button component. 

Also, how would you go about moving the rule to CSS only?

kennyadsl: I mean that we can replace the `-small` definition from the HTML to the CSS. Something like (this is pseudo-code):

```
.coupon-code {
  .text-input {
    // code that make the button small
  }
}
```

In my perspective, this way, making the button smaller is a responsibility of the CSS only. If someone wants to make it bigger, it's a style/look&feel concern, they should not touch the HTML.

dinohamzic: The issue here is that if we need the button smaller in another place (we will be needing it in the checkout process), we would need to duplicate the css is several places.

Passing props that control the style (size, color, visibility etc.) is a common pattern in React for example. It avoids a lot of code duplication.

PR #27, opened 2020-04-10
Update the component structure using dev_support

blocknotes:
This PR is the result of executing `solidus extension` command from dev_support and evaluating the new changes. One of the main feature offered by this upgrade is [sandbox app](https://github.com/solidusio/solidus_dev_support#sandbox-app).


bin/sandbox
```diff
@@ -0,0 +1,83 @@
...
+RUBY
+
+unbundled bundle install --gemfile Gemfile
+
+unbundled bundle exec rake db:drop db:create
+
+unbundled bundle exec rails generate spree:install \
+  --auto-accept \
+  --user_class=Spree::User \
+  --enforce_available_locales=true \
```
kennyadsl, line 73: we need to add `--with-authentication=false` here, see https://github.com/solidusio/solidus_dev_support/pull/96

blocknotes: Done and rebased to remove the last test commit

PR #32, opened 2020-04-17
Refactor - Checkout address step

dinohamzic:
## Description

This is the refactor of:
- the general layout of the checkout process
- the address step in the checkout process
- the order summary section of the checkout process

## Motivation and Context

This is in line with our plan to refactor and modernize all the pages of the Solidus storefront.

## How Has This Been Tested?

All legacy tests pass, small fixes were applied.

Screen Shot 2020-04-17 at 20 06 27

## Types of changes

## Checklist:


app/views/spree/components/forms/inputs/_text.html.erb
```diff
@@ -2,11 +2,14 @@
...
 
   # Optional props
+  autocomplete = local_assigns.fetch(:autocomplete, false)
+  autofocus = local_assigns.fetch(:autofocus, false)
   id = local_assigns.fetch(:id, false)
   disabled = local_assigns.fetch(:disabled, false)
   label = local_assigns.fetch(:label, false)
   min = local_assigns.fetch(:min, false)
-  placeholder = local_assigns.fetch(:placeholder, '')
+  placeholder = local_assigns.fetch(:placeholder, false)
```
kennyadsl, line 11: Why this changed from `''` to `false`?

dinohamzic: Because otherwise the placeholder attribute is rendered with an empty string in the HTML. Ideally we don't want it at all if we don't pass it to the component.

PR #34, opened 2020-04-24
Refactor - Checkout delivery step

dinohamzic:
## Description

This is a refactor of the checkout delivery step.

The main points are:
- refactor of the products list
- refactor of the delivery options and rates
- refactor of the optional delivery message

## Motivation and Context

This is part of the ongoing website-wide refactor initiative (https://github.com/nebulab/solidus_starter_frontend/issues/35).

## How Has This Been Tested?

All the old specs are passing without major changes.

Screen Shot 2020-04-24 at 13 19 21

## Types of changes

## Checklist:


app/views/spree/components/products/_shipment_items.html.erb
```diff
@@ -0,0 +1,22 @@
+<ul class="shipment-items">
+  <% products.each do |product| %>
```
kennyadsl, line 2: I think this variable name should match the file name and should be `shipment_items do |shipment_item|`. I'm also unsure that `components/products` is a good place for this. Maybe it deserves its own namespace since at the end it's not highly related to products.

dinohamzic: Makes sense, will fix soon. Thanks!

dinohamzic: Fixed the naming and moved the file to the checkout folder for the time being. Let me know if you have a better idea in mind.

dinohamzic: Just noticed that I need do move the css files.

PR #42, opened 2020-05-22
Refactor checkout confirm step

dinohamzic:
This is a refactor of the following sections of the application:
- checkout confirmation step
- order complete page
- order page

## Description

The biggest modification is the refactor of the `order_details` component, which is shared by all the pages listed above.

## How Has This Been Tested?

All the old specs are passing with minimal changes.

Screen Shot 2020-05-22 at 15 05 29

Screen Shot 2020-05-25 at 09 53 49

## Types of changes

## Checklist:


app/views/spree/shared/_order_details.html.erb
```diff
@@ -1,147 +0,0 @@
-<% if order.has_checkout_step?("address") %>
```
kennyadsl, line 1: IIRC this is also used by solidus backend.

app/views/spree/components/cart/_item_info.html.erb
```diff
@@ -7,13 +12,15 @@
     <%= variant.options_text %>
   </p>
 
-  <% if line_item.insufficient_stock? %>
+  <% if stock_info && line_item.insufficient_stock? %>
```
kennyadsl, line 15: I think this should probably be another component called line_item_availability, but we can do that in another PR. Can you please just note this somewhere so we don't keep track of it?

PR #44, opened 2020-05-28
Decorate the auth controllers and views for each request

vassalloandrea:
## Description

Replace `after_initialize` with `to_prepare` to prepend the gem 
controllers and views for every request overriding the other extensions.

## Motivation and Context

Using the `after_initialize` method, the application overrides the controllers and the view once.
We have to give the priority to solidus_starter_frontend for each request.

## Types of changes

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,19 +66,19 @@ app.
 
 Simply call the generator to copy all views into your host app:
 ```bash
-$ bundle exec rails g solidus:views:override
+$ bundle exec rails g solidus_starter_frontend:views:override
 ```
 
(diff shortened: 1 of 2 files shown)


jontarg:
@vassalloandrea I'm wondering what's the suggested way to customize the auth views since those in `solidus_starter_frontend` have precedence over the ones in the host app. I just figured out that after trying to customize an empty `app/views/spree/components/navigation/_auth_link.html.erb` in my app with no luck before finding this PR.

Any guidance is appreciated.

PR #55, opened 2020-06-12
Add a solidus compare script

blocknotes:
It will allow us to monitor the changes on a Solidus project path.

## Usage
- To list the changes in diff format: `bin/solidus_compare`
- Apply the required changes
- To update the ignores list (ignore key in _config/solidus_compare.yml_): `bin/solidus_compare -u`

## Description
To keep our code in sync with the latest changes on Solidus frontend we can use this tool to list what changed.

## Types of changes

## Checklist:

### DIFF
--- a/bin/solidus_compare
+++ b/bin/solidus_compare
@@ -8,17 +8,6 @@ require 'time'
 require 'yaml'
 require 'zlib'
 
-PROJECT_PATHS = [
-  'app/controllers',
-  'app/helpers',
-]
-PROJECT_BRANCH = 'master'
-
-SOURCE_REPO = 'https://github.com/solidusio/solidus.git'
-SOURCE_NAME = 'solidus'
-SOURCE_BRANCH = 'master'
-SOURCE_BASE_PATH = 'frontend/'
-
 CONFIG_PATH = 'config/solidus_compare.yml'
 
 @config = {}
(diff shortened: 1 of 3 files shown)


bin/solidus_compare
```diff
@@ -0,0 +1,36 @@
...
+  puts "\n> #{path} changed files"
+  puts list
+  path
+end.compact
+
+if diff.any?
+  path = diff.first
+  cmd = "git diff master:#{path} remotes/#{REMOTE_SOURCE}/master:frontend/#{path}"
+  puts "\n#{cmd}\n  to check the changes"
+end
```
kennyadsl, line 36: What about pushing the output to a temp file, that we can also use to push the output of the diff in the [CircleCI artifacts tab](https://circleci.com/docs/2.0/artifacts/#artifacts-overview)?

blocknotes: Nice idea 👍 
I'm still looking for an easy to read format for differences. But we can start with just a diff a see how it goes

kennyadsl:
This is great, but I don't understand how can we know something changed. Shouldn't the check be red when new changes are detected?

blocknotes:
> This is great, but I don't understand how can we know something changed. Shouldn't the check be red when new changes are detected?

I preferred to avoid blocking the CI workflow, but I'm not completely convinced about this choice currently 😄

PR #56, opened 2020-06-12
Refactor JavaScript v3

dinohamzic:
## Description

This PR removes all the JavaScript for coupon code submit functionality. Instead, when a coupon code is submitted, the page is now refreshed.

Also, we unified the coupon code forms, and we now only have a single component for coupon code submission, instead of 2.

## Motivation and Context

This is part of the current effort to remove jQuery as a dependency and clean up the JavaScript.

## How Has This Been Tested?

All existing tests are passing without modifications.

## Checklist:


spec/system/coupon_code_spec.rb
```diff
@@ -80,45 +80,6 @@ def create_basic_coupon_promotion(code)
           end
         end
       end
-
```
kennyadsl, line 83: I think this needs an explanation (maybe just in the commit description). Why it's ok to remove this spec?

dinohamzic: It's testing 2 unrelated things (coupon code submission and payment method selector) that are being already tested separately in their respective specs. I'll add the explanation in the commit description. 👍

PR #57, opened 2020-06-12
Add responsive setup

aleph1ow:
ref: #59 
- Setup mixins for media queries.
- Apply tablet and/or desktop mixins to get an overall responsive layout.

Each commit addresses a component/section (inspecting every single commit makes the review easier).

| Desktop | Mobile |
|--|--|
|screencapture-localhost-3000-t-brand-solidus-2020-06-26-19_33_59|screencapture-localhost-3000-t-brand-solidus-2020-06-26-19_34_07|
|screencapture-localhost-3000-cart-2020-06-26-19_34_59|screencapture-localhost-3000-cart-2020-06-26-19_35_09|
|screencapture-localhost-3000-checkout-address-2020-06-26-19_35_32|screencapture-localhost-3000-checkout-address-2020-06-26-19_35_47|
|screencapture-localhost-3000-checkout-confirm-2020-06-26-19_36_09|screencapture-localhost-3000-checkout-confirm-2020-06-26-19_36_18|

## Types of changes

## Checklist:

### DIFF
--- a/app/assets/stylesheets/spree/frontend/base/_responsive.scss
+++ b/app/assets/stylesheets/spree/frontend/base/_responsive.scss
@@ -1,0 +1,21 @@
+$tablet-screen: 600px;
+$small-screen: 900px;
+$wide-screen: 1150px;
+
+@mixin tablet {
+  @media (min-width: $tablet-screen) {
+    @content;
+  }
+}
+
+@mixin desktop {
+  @media (min-width: $small-screen) {
+    @content;
+  }
+}
+
+@mixin wide-desktop {
+  @media (min-width: $wide-screen) {
+    @content;
+  }
+}
--- a/app/assets/stylesheets/spree/frontend/components/layout/_header.scss
+++ b/app/assets/stylesheets/spree/frontend/components/layout/_header.scss
@@ -16,7 +16,46 @@
     &__content {
       align-items: center;
       display: flex;
+      flex-wrap: wrap;
+      justify-content: space-between;
       padding: 2rem 3rem;
+
+      @include desktop {
+        flex-wrap: nowrap;
+        justify-content: flex-end;
+      }
+
+      .logo {
+        order: 1;
+
+        img {
+          width: 10em;
+
+          @include tablet {
+            width: auto;
+          }
+        }
+      }
+
+      .search-bar {
+        flex: 0 0 100%;
+        margin-top: 10px;
+        order: 4;
+
+        @include desktop {
+          flex: 1 0 auto;
+          margin-top: 0;
+          order: 2;
+        }
+      }
+
+      .auth-link {
+        order: 2;
+      }
+
+      .cart-link {
+        order: 3;
+      }
     }
   }
 }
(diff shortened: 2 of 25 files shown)


app/assets/stylesheets/spree/frontend/base/_responsive.scss
```diff
@@ -0,0 +1,21 @@
+$tablet-screen: 660px;
```
dinohamzic, line 1: Is this a standard tabled width? I remember `768` being a common one, but I could be wrong.

app/assets/stylesheets/spree/frontend/components/layout/_header.scss
```diff
@@ -16,7 +16,46 @@
...
+      flex-wrap: wrap;
+      justify-content: space-between;
       padding: 2rem 3rem;
+
+      @include desktop {
+        flex-wrap: nowrap;
+        justify-content: flex-end;
+      }
+
+      .logo {
```
dinohamzic, line 28: We should move this styles to `_logo.scss`. Is there a reason why they are here?

app/assets/stylesheets/spree/frontend/components/layout/_header.scss
```diff
@@ -16,7 +16,46 @@
...
+        img {
+          width: 10em;
+
+          @include tablet {
+            width: auto;
+          }
+        }
+      }
+
+      .search-bar {
```
dinohamzic, line 40: We should move this styles to `_search_bar.scss`. Is there a reason why they are here?

app/assets/stylesheets/spree/frontend/components/layout/_header.scss
```diff
@@ -16,7 +16,46 @@
...
+        order: 4;
+
+        @include desktop {
+          flex: 1 0 auto;
+          margin-top: 0;
+          order: 2;
+        }
+      }
+
+      .auth-link {
```
dinohamzic, line 52: We should move this styles to `_auth_link.scss`. Is there a reason why they are here?

app/assets/stylesheets/spree/frontend/components/layout/_header.scss
```diff
@@ -16,7 +16,46 @@
...
+        order: 2;
+      }
+
+      .cart-link {
```
dinohamzic, line 56: We should move this styles to `_cart_link.scss`. Is there a reason why they are here?

app/assets/stylesheets/spree/frontend/base/_responsive.scss
aleph1ow: 768px is a standard but some tablets starts even at 600px.

app/assets/stylesheets/spree/frontend/components/layout/_header.scss
aleph1ow: I think that I get a similar comment before, regarding where to put style rules that belongs to the aggregation of components. For example, having the `order` inside the single component does not make sense to me because the same component could have a different order if it's used in another aggregation. In this case, the header aggregates a logo, search-bar, auth and cart links and sort them according to mobile/desktop layout.

app/assets/stylesheets/spree/frontend/base/_responsive.scss
aleph1ow: Maybe I could lower it to 600px instead.

dinohamzic: If the layout works at that size go for it 👍

PR #58, opened 2020-06-12
Fix Login Form

aleph1ow:
Restore the following functionalities, passing the right form field name to the text input partial:
- login
- signup
- password reset
- guest checkout

## Motivation and Context

Having single form fields now refactored using generic partials, we must pass the exact form field name to the partial. The original view was using a form helper and the field name was generated by the form builder object. It was an oversight.

## Types of changes

## Checklist:


dinohamzic:
Nice, login works! 🎉

Should we do this also for the other forms (registration, guest login, password reset)?

aleph1ow:
@dinohamzic yes I think so...going to check!

PR #64, opened 2020-07-09
Fix gateway partial that uses non-existing `billing_name` method

jontarg:
## Description

Use the two name methods available instead.

`billing_name` doesn't seem to be a valid method, at least on Solidus v2.10 (the one I'm using). 

## How Has This Been Tested?

The checkout is broken on a fresh install of this gem using Solidus v2.10.

### DIFF
--- a/app/views/spree/checkout/payment/_gateway.html.erb
+++ b/app/views/spree/checkout/payment/_gateway.html.erb
@@ -7,7 +7,7 @@
     id: "name_on_card_#{payment_method.id}",
     label: t('spree.name_on_card'),
     name: "#{param_prefix}[name]",
-    value: @order.billing_name
+    value: "#{@order.billing_firstname} #{@order.billing_lastname}"
   ) %>
 
   <%= render(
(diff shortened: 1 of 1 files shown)


app/views/spree/checkout/payment/_gateway.html.erb
```diff
@@ -7,7 +7,7 @@
     id: "name_on_card_#{payment_method.id}",
     label: t('spree.name_on_card'),
     name: "#{param_prefix}[name]",
-    value: @order.billing_name
+    value: "#{@order.billing_firstname} #{@order.billing_lastname}"
```
kennyadsl, line 10: We were thinking that this extension would support Solidus starting from v2.11, which is the current master. It would be awesome to make it work with v2.10 as well though, at least until we release v2.11. What about supporting both versions conditionally using one or the other version checking if `@order` responds to `billing_name`?

MFRWDesign: Sounds good to me, and yeah, we'd love it to support v2.10 -- let us know how we can help.

PR #90, opened 2020-09-04
Convert controller specs to requests

SamuelMartini:
## Description
Ref #18 

In converting the controller in request specs I have tried to not stubbing any method. The purpose of this PR is to port every test that was written in the controller and for this reason, it was not possible to avoid stubs in every example.

`CheckoutController` and `OrderController` are the more sensitive tests. `CheckoutController` has many `before_action` filters that make it cumbersome to ensure the example was testing the right filter (when the expectation is the same). Those tests probably make sense in a Controller test where you can stub messages to be received but in a request spec we are only interested in the response so we may want to revisit those tests. Also, It would be really cool to refactor the controller and 
[make it lean](https://github.com/nebulab/solidus_starter_frontend/blob/826359442a4308b5177fcd3d0f1504e39d81e22e/app/controllers/spree/checkout_controller.rb#L9-L23).


kennyadsl:
@SamuelMartini I've opened https://github.com/solidusio/solidus/issues/3873 to track what we need to do in core to remove the hack we did to test system specs without forgery protection.

kennyadsl:
@SamuelMartini forgery is now false on the dummy app in core!

kennyadsl:
@SamuelMartini can you please rebase now that specs are green on master? Also, I'd love to take advantage of this PR to get rid of the `_controller` part in the filename of specs. Now that they are requests specs it doesn't make a lot of sense anymore.

kennyadsl (approved):
Thanks @SamuelMartini for never give up on this huge task!

PR #91, opened 2020-09-04
Update the font-family, adjust the typo

mfrecchiami:
This PR would introduce a general improvement for the typography: the font-family has been updated with the official font used by Solidus (backend and website) since the past year ([Inter Font](https://rsms.me/inter/)).

## What is changed

- Font family
- Font sizes, line-height, margins
- There's a new variables file dedicated to the typography
- On the product page, the price was moved under the product name/title

## Next steps

- Improve the responsiveness
- Add an appealing image to the hero

The aim is to have a more catchy storefront in the near future; improvements made in this PR want to be a solid basis for the next restyling.

## Screenshots

Sample-Store (2)

Categories-Mugs-Sample-Store

Solidus-T-Shirt-Sample-Store

Shopping-Cart-Sample-Store


kennyadsl:
@mfrecchiami beside the compare check, there are actual tests failing here. Do you need any help figuring out what's happen?

davidedistefano:
@mfrecchiami I know you're going to improve the homepage's hero in the next PR, but can we update its copy with the current one on the website? I think it's important to have consistent communication across all Solidus' channels. 

I mean this one:

> The only eCommerce platform you’ll ever need.
> Build, customize and scale your store with no limits or license fees. Solidus is the free, open-source eCommerce framework for digitally-native brands, fast-growing online businesses and pragmatic developers.

Also, probably not related to this PR, but I find the info section on the PDP a bit too long, and the "Add to cart button" far below the fold. Can we move the product properties under the CTA or under the product images?  

Great job as always by the way!

mfrecchiami:
@kennyadsl I've tried to keep the defaults of `scss-lint` but it seems pretty impossible doing something good also for the small devices, keeping the max nesting depth to 3.
I added just two rules to the `scss-lint.yml`, one fix the nesting depth to 6 (might be too much for now, but should be optimal for the future restyling); the other one let us using the vendor prefix.
Generally we don't need them if are using Autoprefixer or similar, but (especially in this PR) we should care about the anti-aliasing of the font rendered.
The `font-smooth` property controls it.
(Just for notice the difference try to disable the rule - it's applied to the `html`)

mfrecchiami:
@davidedistefano thanks for your feedback 🙏 

I updated the message of the hero, and moved the Properties table, under the CTA. This is a faster solution.
We can evaluate to do something more on the PDP in another PR, IMO.

davidedistefano:
@mfrecchiami Agreed. Thanks 🙏

app/views/spree/components/products/_product_taxons.html.erb
```diff
@@ -1,13 +1,13 @@
...
     <h2 class="product-taxons__title">
-      <%= t('spree.look_for_similar_items') %>
+      <%= t('spree.look_for_similar_items') %>:
     </h2>
 
     <ul class="product-taxons__list">
       <% @product.taxons.each do |taxon| %>
         <li>
-          <%= link_to taxon.name, seo_url(taxon) %>
+          <%= link_to taxon.name, seo_url(taxon), class: "link-accent-color" %>
```
kennyadsl, line 10: I'd avoid these `link-accent-color` classes in favor of adding this style in the CSS. If we want to keep a class in the HTML it should have a semantic meaning. Why does this link have a different style? If it's an appearance concern, HTML should not be aware of that.

kennyadsl (changes requested):
Just found one thing that I think we should address to be consistent with the way we structured HTML/CSS in the project. Other than that, this works for me, thanks @mfrecchiami!

mfrecchiami:
@kennyadsl you were right: it wasn't worth to keep that class for just 3 "secondary actions". I probably made it before adjusted the strict parameters that linter had!
Thank you for your feedback 🙏

kennyadsl (approved):
👍

PR #95, opened 2020-09-07
Check if localeSelector exists before using it

### LINKED ISSUE #94: JS error related to locale when the app has a single locale available
With a single locale in the store, I get this console error:
```
locale_selector.source.js:3 Uncaught TypeError: Cannot read property 'addEventListener' of null
    at locale_selector.source.js:3
```

Adding another locale to `available_locales` of the current store makes the error disappear.

blocknotes:
When the application has a single locale set the locale selector element
is not present.

Closes #94 

## Types of changes

## Checklist:

### DIFF
--- a/app/assets/javascripts/spree/frontend/locale_selector.js
+++ b/app/assets/javascripts/spree/frontend/locale_selector.js
@@ -1,7 +1,9 @@
 window.addEventListener('DOMContentLoaded', () => {
   const localeSelector = document.querySelector('.locale-selector select');
 
-  localeSelector.addEventListener('change', (event) => {
-    event.target.form.submit();
-  });
+  if(localeSelector) {
+    localeSelector.addEventListener('change', (event) => {
+      event.target.form.submit();
+    });
+  }
 });


kennyadsl (commented):
Why also removing all the `;`?

blocknotes:
> Why also removing all the `;`?

Uhm, I'm used to removing them (it will be handled by the JS processor, as the arrow function or any ES6 code).

Do you think it's better to keep them to maintain the same appearance of the other JS or I can leave it as is?

kennyadsl:
I'd like to be consistent and I'm in favor of the refactor if it makes sense, but I think it should be done in a specific task, unrelated to this PR, after a discussion with the team.

blocknotes:
> I'd like to be consistent and I'm in favor of the refactor if it makes sense, but I think it should be done in a specific task, unrelated to this PR, after a discussion with the team.

I fixed up the last commit to restoring the `;`

kennyadsl (approved):
Thanks!

PR #103, opened 2020-09-30
Fix the CI

elia:
Fixed the CI after recent changes in solidus_dev_support and solidus-master broke it.

## Description

I avoided adding the lint job to the CI because I expect it to need much more work that is better handled in a separate PR.
The same goes for keeping apparition instead of migrating to webdrivers, it would require too many changes and the priority is to have a green CI.

## How Has This Been Tested?

```
bin/setup
bin/rake
```

## Types of changes

## Checklist:


bin/sandbox
```diff
@@ -72,9 +72,11 @@ unbundled bundle exec rails generate spree:install \
   --user_class=Spree::User \
   --enforce_available_locales=true \
   --with-authentication=false \
+  --payment-method=none \
   $@
 
 unbundled bundle exec rails generate solidus:auth:install
+unbundled bundle exec rails generate ${extension_name}:install
```
kennyadsl, line 79: Is this supposed to be interpolated?

kennyadsl (commented):
Found a little thing that could be wrong but thanks for fixing this!

bin/sandbox
elia: Yes, that's coming from dev-support and it's valid bash syntax, although less common

PR #112, opened 2020-10-09
Update Sandbox's Gemfile to include Solidus components

### LINKED ISSUE #111: Sandbox seems broken
When I create the sandbox I get this error:
```
/projects/.asdf/installs/ruby/2.7.1/lib/ruby/gems/2.7.0/gems/actionpack-6.0.3.4/lib/action_dispatch/routing/route_set.rb:578:in `add_route': Invalid route name, already in use: 'root'  (ArgumentError)
You may have defined two routes with the same name using the `:as` option, or you may be overriding a route already defined by a resource with the same naming. For the latter, you can restrict the routes created with `resources` as explained here:
https://guides.rubyonrails.org/routing.html#restricting-the-routes-created
```

#### Steps to reproduce
- Clone the repo: `git clone https://github.com/nebulab/solidus_starter_frontend`
- Enter in the project root: `cd solidus_starter_frontend`
- Run bundle: `bundle`
- Launch the sandbox executable: `bin/sandbox`

blocknotes:
Replace solidus gem, as exposed by the README.

Closes #111.

## Types of changes

## Checklist:

### DIFF
--- a/bin/sandbox
+++ b/bin/sandbox
@@ -49,18 +49,15 @@ fi
 
 cd ./sandbox
 cat <<RUBY >> Gemfile
-gem 'solidus', github: 'solidusio/solidus', branch: '$BRANCH'
-gem 'solidus_auth_devise', '>= 2.1.0'
-gem 'rails-i18n'
-gem 'solidus_i18n'
-
+# By default, the `solidus` gem also includes the standard frontend via
+# the `solidus_frontend` gem. To make this extension work, you need to
+# exclude it and manually include all the other Solidus components.
+gem 'solidus_core'
+gem 'solidus_api'
+gem 'solidus_backend'
+gem 'solidus_sample'
 gem '$extension_name', path: '..'
-
-group :test, :development do
-  platforms :mri do
-    gem 'pry-byebug'
-  end
-end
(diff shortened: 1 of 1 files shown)


bin/sandbox
```diff
@@ -49,18 +49,12 @@ fi
 
 cd ./sandbox
 cat <<RUBY >> Gemfile
-gem 'solidus', github: 'solidusio/solidus', branch: '$BRANCH'
-gem 'solidus_auth_devise', '>= 2.1.0'
-gem 'rails-i18n'
-gem 'solidus_i18n'
-
```
kennyadsl, line 56: I think this error has been introduced by the script that tries to copy the extensions' template in order to update the extensions. If we add a comment here that explain that the list of individual gems has a specific purpose, it will less likely happen again.

kennyadsl (approved):
Left a possible improvement to avoid regressions. Up to you if it's worth it.

bin/sandbox
blocknotes: Good point 👍

blocknotes: Force-pushed an update ✅

PR #121, opened 2020-10-30
Fix missing translations spec module

blocknotes:
The conditional require (on Solidus version check > 2.11) is loading the
file for missing translations check but the module must be loaded in RSpec
config to avoid the method missing.

Perhaps we could propose a PR to Solidus to include the module implicitly in the translations file: [here](https://github.com/solidusio/solidus/blob/master/core/lib/spree/testing_support/translations.rb#L17).

## Types of changes

## Checklist:


kennyadsl (approved):

blocknotes:
Proposed as draft PR for now: https://github.com/solidusio/solidus/pull/3821

WDYT @kennyadsl?

PR #122, opened 2020-10-30
Add the solidus_i18n gem to the sandbox

### LINKED ISSUE #119: Locale selector missing in Sandbox
For some reason, the locale selector disappeared in the restored Sandbox.
See `app/views/spree/components/navigation/_locale_selector.html.erb`
`current_store.available_locales` just contains [:en] language.
When using the dummy app the selector is present and `current_store.available_locales` is fully populated with all language choices.

ChristianRimondi:
Closes #119

## Description

This PR adds the gem [solidus_i18n](https://github.com/solidusio/solidus_i18n) to the sandbox.

## Motivation and Context

The gem is present in the dummy app but was missing from the sandbox. Its presence is required to show the locale selector in the storefront.

## How Has This Been Tested?

I've run the sandbox using [this guide](https://github.com/nebulab/solidus_starter_frontend/blob/master/docs/development.md#running-the-sandbox) making sure that the locale selector was visible.

## Types of changes

## Checklist:

### DIFF
--- a/bin/sandbox
+++ b/bin/sandbox
@@ -56,6 +56,8 @@ gem 'solidus_core'
 gem 'solidus_api'
 gem 'solidus_backend'
 gem 'solidus_sample'
+gem 'rails-i18n'
+gem 'solidus_i18n'
 gem '$extension_name', path: '..'
 gem 'solidus_auth_devise'
 RUBY


blocknotes (approved):

bin/sandbox
```diff
@@ -58,6 +58,7 @@ gem 'solidus_backend'
 gem 'solidus_sample'
 gem '$extension_name', path: '..'
 gem 'solidus_auth_devise'
+gem 'solidus_i18n'
```
kennyadsl, line 61: This file is generated/updated with the `solidus extension .` command if I recall correctly. If we run it again in the future it will try to remove this line and we'll probably forget that having it here was intentional. What about adding a small comment to explain that this is actually needed?

ChristianRimondi: I just noticed that `solidus_i18n` is already present in [solidus_dev_support](https://github.com/solidusio/solidus_dev_support/blob/master/lib/solidus_dev_support/templates/extension/bin/sandbox.tt#L55) so there shouldn't be any problem if we run that command again in the future. However, I'm not sure if I need to also add `rails_i18n` 🤔 . WDYT?

Issue #79, opened 2020-07-31
Check the components behaviour with cache server enabled


blocknotes:
An example is in categories page using `bin/rails dev:cache`.
After the first visit the partial "Shop by Categories" seems to be cached and the current category is not highlighted anymore.

Image:
![image](https://user-images.githubusercontent.com/6893256/89025225-c07f2f00-d326-11ea-9b5f-bb9ea1e253c2.png)

Without cache the current category is in bold.

fthobe:
@jarednorman can be closed

PR #127, opened 2020-11-13
Fix cache issue in sidebar

blocknotes:
Related to #79, and hopefully closes it - but I would ask more eyes here for more checks, it could be that I missed something.

## Description
- When cache is enabled, the taxonomies current element is not updated because the outside cache key is not considering the current taxon.

## How Has This Been Tested?
- Enabled cache: `bin/rails dev:cache`
- Went to the categories section
- Switched categories

## Types of changes

## Checklist:

### DIFF
--- a/app/views/spree/components/navigation/_taxonomies.html.erb
+++ b/app/views/spree/components/navigation/_taxonomies.html.erb
@@ -1,7 +1,7 @@
 <% max_level = Spree::Config[:max_level_in_taxons_menu] || 1 %>
 
 <nav class="taxonomies">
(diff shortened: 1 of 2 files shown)


kennyadsl (approved):
Makes sense, Maybe we can add a quick test for this to be sure we won't break navigation again in the future.

ChristianRimondi (approved):

blocknotes:
> Makes sense, Maybe we can add a quick test for this to be sure we won't break navigation again in the future.

Sure.
Added a fixup commit with a test.

PR #142, opened 2021-01-12
Add source_code_uri to gemspec

kennyadsl:
To please octokit while autogenerating the changelog


kennyadsl:
I was using an outdated version, this is already fixed.

PR #145, opened 2021-02-18
Fix the CI 

kennyadsl:
## Description

Solidus master now tracks 3.0 and we are testing this gem against it. Almost all the deprecated code was removed but there was still something.


kennyadsl:
I still couldn't figure out what happens with specs here. My initial investigation showed that for [some specs](https://github.com/nebulab/solidus_starter_frontend/blob/54dd381738540bd882e2809a8cb0741055cc2bfe/spec/controllers/spree/checkout_controller_spec.rb#L301-L329), the [`ensure_inventory_units`](https://github.com/solidusio/solidus/blob/dc813e6a69b9b2074d9eb0e92486980ad5ff23b6/core/app/models/spree/order.rb#L757-L764) (called from [here](https://github.com/solidusio/solidus/blob/a17109b07a817ae34bc61491c3ed1ea824ee32c5/core/lib/spree/core/state_machines/order.rb#L117)) raise `InsufficientStock`, even if there's apparently nothing in the specs that could make it behave this way. 

Also, we have the same exact tests in [solidus_frontend](https://github.com/solidusio/solidus/blob/a17109b07a817ae34bc61491c3ed1ea824ee32c5/frontend/spec/controllers/spree/checkout_controller_spec.rb#L301-L329) and they work correctly. 

🤔

PR #148, opened 2021-03-01
Update test to use discard instead of paranoia

waiting-for-dev:
Ref solidusio/solidus@b0092d0f3

## Description
The Master branch on solidus is not using paranoia anymore, so we need to update from our end.

It wasn't caught by the CI because, if I'm not wrong, it's using released solidus versions.

## Types of changes

## Checklist:

### DIFF
--- a/spec/system/checkout_spec.rb
+++ b/spec/system/checkout_spec.rb
@@ -336,7 +336,7 @@ describe 'Checkout', type: :system, inaccessible: true do
   end
 
   context "user has payment sources", js: true do
-    before { Spree::PaymentMethod.all.each(&:really_destroy!) }
+    before { Spree::PaymentMethod.all.each(&:destroy!) }
     let!(:bogus) { create(:credit_card_payment_method) }
     let(:user) { create(:user) }
 


kennyadsl:
@waiting-for-dev I think this commit is already in https://github.com/nebulab/solidus_starter_frontend/pull/145. I know that @SamuelMartini also did find why the remaining specs are failing. @SamuelMartini can you please report here or in #145 what you discovered?

SamuelMartini:
> @waiting-for-dev I think this commit is already in #145. I know that @SamuelMartini also did find why the remaining specs are failing. @SamuelMartini can you please report here or in #145 what you discovered?

The error I'm referring to is 
```ruby
NoMethodError:
  undefined method `name' for nil:NilClass
```
That currently happens in `spec/system/products_spec.rb:164` and `/spec/system/products_spec.rb:198`.
This is rails trying to detect the [active storage service name](https://github.com/rails/rails/blob/66b861f24cd8ca34b726f77b46f6b25b21f0b972/activestorage/app/models/active_storage/blob.rb#L52)
which is not set in the dummy app the tests are using.

There are 2 dummy apps that solidus can create:

1. Dummy::Application to test extensions that use this [DummyGenerator#test_dummy_config](https://github.com/solidusio/solidus/blob/master/core/lib/generators/spree/dummy/dummy_generator.rb#L53). This generator creates a full rails application (Dummy::Application) also used by solidus_starter_frontend.
2. DummyApp::Application to test the solidus framework and it is an application initialized in-memory [here](https://github.com/solidusio/solidus/blob/v2.11.4/core/lib/spree/testing_support/dummy_app.rb#L35).

[This commit](https://github.com/solidusio/solidus/commit/c01d6ac85a1f98789eb6e4d13660041b4dac7cba) configures the active storage for the solidus DummyApp::Application(2).

We need to add a similar configuration to the DummyGenerator(1) so that extensions(and solidus_starter_frontend) that run on solidus 3 have the service set.

waiting-for-dev:
Then, probably this PR can be closed  :slightly_smiling_face:

PR #153, opened 2021-03-25
Include Devise routes/controllers/views directly in the project

kennyadsl:
## Description

This PR explores if it makes still sense to have all the views needed for the frontend copied over the host application except for the things related to the authentication. The rationale is that if people want to extend their application they probably also want to change those parts of the application.

Also, it was technically very difficult to make SolidusAuthDevise work without all those things copied because it requires its
components (controller, routes, views) only when SolidusSupport.frontend_available? returns true. This method returns false using this frontend, even using a monkey_patch, because some calls are made during the engines initialization, when the decorators are not yet loaded.

**Cons** of this approach are:

1. views/controllers/routes are duplicated here and in solidus_auth_devise
2. at this point it's not possible to install `solidus_starter_fronted` without `solidus_auth_devise`. Well, it's possible but al the devise things need to be manually removed from the host application after the installation.

## Motivation and Context

We want to provide a smooth installation process, and now it was impossible to have a store working with the auth provided by solidus_auth_devise.

The project will ship with a working Authentication system that uses Devise. Don't you like it? Well, you can just remove the extra code from your project after it has been copied and use what you prefer.

With another iteration, we can explore adding an install configuration (`--without-authentication`) that does not copy auth-related things if needed.

## How Has This Been Tested?

I tested this creating a new rails store following the README instructions updated with this PR.
Some specs are passing, some don't but that's because we need to change some things that solidus_auth_devise views changed, but it's expected.

## Types of changes

## Checklist:

- [ ] Fix current specs
- [ ] Add specs from SolidusAuthDevise
- [ ] Fix TODOs introduced
- [ ] Update solidus_compare step hash


waiting-for-dev:
> Also, it was technically very difficult to make SolidusAuthDevise work without all those things copied because it requires its
components (controller, routes, views) only when SolidusSupport.frontend_available? returns true. This method returns false using this frontend, even using a monkey_patch, because some calls are made during the engines initialization, when the decorators are not yet loaded.

We are making available two methods to install `solidus_starter_frontend`:

**1. Copying all `solidus_starter_frontend` app files over the project, through `solidus_starter_frontend` gem's executable.**

In this case, the executable relies on `rails generate solidus:install` having been already executed. This generator is responsible for adding `solidus_auth_devise` to the stack. At the time it gets installed, the frontend is not available (we haven't executed `solidus_starter_frontend` yet) so `solidus_auth_devise` doesn't add the frontend stuff:

https://github.com/solidusio/solidus_auth_devise/blob/09841c164e1d0e6ab7afbb80494737fc692eb806/lib/spree/auth/engine.rb#L26

**2. Adding `solidus_starter_frontend` gem to the `Gemfile`.**

For this installation option, the README is currently recommending adding `solidus_starter_frontend` **after** `bin/rails generate solidus:install` has been executed. So, it essentially happens the same as in option 1, but on top of devise resources not being available an error is raised because of this references in `solidus_starter_frontend` initializer:

https://github.com/nebulab/solidus_starter_frontend/blob/1609b364d670972bbba718b44e406729d25242f7/lib/solidus_starter_frontend/engine.rb#L18-L25

(`Spree::Auth::Engine` is defined, but it doesn't have created the `Controller`'s classes.)

However, if we add `solidus_starter_frontend` **before** executing `bin/rails generate solidus:install`, i.e. at the same time that we add `solidus_auth_devise`, `solidus_core`, etc... to the Gemfile, the installation does succeed and devise routes are found.

What I'd propose to fix the problem is:

1. Fix instructions for Option 2.
2. Remove Option 1. Yeah, it sounds drastic. However, if we take a step backward, I think that supporting an installation method where everything gets copied is not a good idea. It lacks any kind of modularity. Instead, we could add a task that allows you to copy only what you need to override, in a way similar to what devise provides. E.g.:

```bash
rails generate solidus_starter_frontend:controllers [checkout]
```

It would provide a better user experience and, on top of that, this issue would be fixed.

kennyadsl:
@waiting-for-dev I'm fine trying the proposed approach! We can always use a generator to copy controllers and everything else by default if needed. 👍 Feel free to upset the installation process!

gsmendoza:
@kennyadsl I'm closing this PR since it is now superseded by https://github.com/nebulab/solidus_starter_frontend/pull/172.

PR #157, opened 2021-04-29
Update gemspec to allow Solidus 3

kennyadsl:
## Description

This is needed because the last version of Solidus is 3.0


spaghetticode (approved):
❤️ 3.0

Issue #161, opened 2021-06-04
Auth issue when getting started


alexisagos:
Hi,

I'd like to report issues that I'm encountering with solidus_starter_frontend.
I understand this is the recommended route for newcomers but in my case getting started with it hasn't been very simple.
I have outlined the steps I followed hoping we can fix this and make it easier to get started with the new frontend.
I'm still investigating to get past this issue, any help is very welcome !

With `ruby 2.7.2p137 (2020-10-01 revision 5445e04352) [x86_64-darwin19]`and `Rails 6.1.3.2`, if we create a fresh rails app `rails new test_solidus --skip-javascript`:

Add the following gems to the Gemfile:
```
gem "solidus_core"
gem "solidus_api"
gem "solidus_backend"
gem "solidus_starter_frontend"
```

Run the installer `rails generate solidus:install`.

Serve the app `rails s`.

Then the app works but the styling is missing, login seems to work too..
<img width="1439" alt="Screenshot 2021-06-04 at 11 07 26" src="https://user-images.githubusercontent.com/17249120/120777922-195ebb80-c526-11eb-917e-dcf733d48b11.png">
<img width="1440" alt="Screenshot 2021-06-04 at 11 07 37" src="https://user-images.githubusercontent.com/17249120/120777931-1a8fe880-c526-11eb-8fbc-fc345586be41.png">

The auth controllers are present if we check the console `rails c`:
```
Running via Spring preloader in process 3709
Loading development environment (Rails 6.1.3.2)
irb(main):001:0> Spree::UserSessionsController
=> Spree::UserSessionsController
```

So if we want to go the other way and add the files to the project `bundle exec solidus_starter_frontend`, comment out the solidus_starter_frontend gem in the Gemfile:
```
gem "solidus_core"
gem "solidus_api"
gem "solidus_backend"
# gem "solidus_starter_frontend"
```

After `bundle install` and `rails s`:
<img width="1430" alt="Screenshot 2021-06-04 at 11 20 46" src="https://user-images.githubusercontent.com/17249120/120779045-f5e84080-c526-11eb-89ea-c0b722b7d82b.png">

Now, we get a nice store, everything looks good but no auth link in the top bar and auth doesn't work.
If we navigate to "/login", we get `No route matches [GET] "/login"`

If we check the console `rails c`:
```
Running via Spring preloader in process 4442
Loading development environment (Rails 6.1.3.2)
irb(main):001:0> UserSessionsController
[...]

PR #162, opened 2021-06-11
Fix installation method as a gem executable

waiting-for-dev:
This is a WIP that depends on changes that need to be integrated on [`solidus_support`](https://github.com/solidusio/solidus_support). In order to try it:

- Check out this branch or your local repository.
- Make sure you don't have other installation of `solidus_starter_frontend`: `gem uninstall solidus_starter_frontend`
- Install the gem from this branch:
  - `gem build solidus_starter_frontend.gemspec`
  - `gem install solidus_starter_frontend-0.1.0.gem`
- Create a rails app: `rails new store --skip-javascript && cd store`
- Add solidus gems: `bundle add solidus_core solidus_backend solidus_api solidus_sample`
- **IMPORTANT**: Add `gem 'solidus_support', github: 'waiting-for-dev/solidus_support', branch: 'waiting-for-dev/change_engine_detection'` to the `Gemfile`
- Run `bundle`
- Run `bin/rails g solidus:install`
- Say `y` to the prompt suggesting `solidus_auth_devise`
- Run `solidus_starter_frontend`
- Run rails server `bin/rails server`
- Go to `http://localhost:3000` and check that the devise routes are working. E.g., `http://localhost:3000/login`.

## Description

To support `solidus_auth_devise`, it needs to rely on a new version of
`solidus_support` that changes how it detects if a Solidus frontend is
available and the moment it adds the extensions' controller & view
directories to the Rails known paths.

## Motivation and Context

The previous `solidus_support` version added the extensions' paths on
the engine's (in this case, `solidus_auth_devise`'s engine) load time
and only when `SolidusSupport#frontend_available?` returned `true`.
Engines are loaded before the application code, so at that time, the
monkey patching of `#frontend_available?` that we were generating was
not at hand. The solution is two-fold: we add the paths on an
`initializer` on one side. However, at that moment, the application code
is neither available, so on the other side, we change
`#frontend_available?` to look for a
`Rails.configuration.x.solidius.frontend_available` setting that we add
to `application.rb` using the generator code. To understand the whole
picture, keep in mind that `Rails.configuration` is available on an
initializer, but it's not on an engine's load time.
On top of that, we need to update a couple of things on the generated code:

- We need to reference `Rails.root` instead of `Engine.root`, as
  `SolidusStarterFrontend::Engine` is unavailable on this installation
  method.
- We need to copy to `application.rb` the custom code on
  `SolidusStarterFrontend::Engine` for the same reason as above.

## How Has This Been Tested?
Tested the installation method locally.

## Types of changes

## Checklist:

References #153 & #161

### DIFF
--- a/lib/generators/solidus_starter_frontend/solidus_starter_frontend_generator.rb
+++ b/lib/generators/solidus_starter_frontend/solidus_starter_frontend_generator.rb
@@ -5,26 +5,39 @@ class SolidusStarterFrontendGenerator < Rails::Generators::Base
 
   def install
     # Copy directories
-    directory 'app', 'app'
-    directory 'lib/views', 'lib/views'
+    directory 'app', 'app', exclude_pattern: /auth_views/
+    directory 'lib/views/auth', 'app/views'
 
     # Copy files
     copy_file 'lib/solidus_starter_frontend_configuration.rb', 'lib/solidus_starter_frontend_configuration.rb'
     copy_file 'lib/solidus_starter_frontend/config.rb', 'lib/solidus_starter_frontend/config.rb'
-    copy_file 'lib/solidus_starter_frontend/solidus_support_extensions.rb', 'lib/solidus_starter_frontend/solidus_support_extensions.rb'
-
-    # Initializer
-    initializer 'solidus_starter_frontend.rb' do
-      "require 'solidus_starter_frontend/solidus_support_extensions'"
-    end
 
     # Routes
     copy_file 'config/routes.rb', 'tmp/routes.rb'
     prepend_file 'config/routes.rb', File.read('tmp/routes.rb')
 
+    # Enable Solidus frontend
+    application do
+      <<~RUBY
+        config.x.solidus.frontend_available = true
+        config.to_prepare do
+          if defined?(Spree::Auth::Engine)
+            [
+              Spree::UserConfirmationsController,
+              Spree::UserPasswordsController,
+              Spree::UserRegistrationsController,
+              Spree::UserSessionsController,
+              Spree::UsersController
+            ].each do |auth_controller|
+              auth_controller.include SolidusStarterFrontend::Taxonomies
+            end
+          end
+        end
+        RUBY
+    end
+
     # Gems
     gem 'canonical-rails'
-    gem 'solidus_support'
     gem 'truncate_html'
 
     # Text updates
--- a/lib/solidus_starter_frontend/engine.rb
+++ b/lib/solidus_starter_frontend/engine.rb
@@ -9,6 +9,10 @@ module SolidusStarterFrontend
 
     engine_name 'solidus_starter_frontend'
 
+    initializer 'solidus_starter_frontend', before: 'solidus_support_frontend_paths' do
+      Rails.configuration.x.solidus.frontend_available = true
+    end
+
     # use rspec for tests
     config.generators do |g|
       g.test_framework :rspec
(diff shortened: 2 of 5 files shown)


app/controllers/concerns/solidus_starter_frontend/auth_views.rb
```diff
@@ -10,7 +10,8 @@ module AuthViews
     protected
 
     def configure_views
-      prepend_view_path Engine.root.join('lib', 'views', 'auth')
+      root = defined?(SolidusStarterFrontend::Engine) ? SolidusStarterFrontend::Engine.root : Rails.root
+      prepend_view_path root.join('lib', 'views', 'auth')
```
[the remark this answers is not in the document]
waiting-for-dev: Good point!! I just updated it. I also had to add the `exclude_pattern` option in:

```ruby
directory 'app', 'app', exclude_pattern: /auth_views/
```

lib/generators/solidus_starter_frontend/solidus_starter_frontend_generator.rb
```diff
@@ -5,26 +5,39 @@ class SolidusStarterFrontendGenerator < Rails::Generators::Base
 
   def install
     # Copy directories
-    directory 'app', 'app'
-    directory 'lib/views', 'lib/views'
+    directory 'app', 'app', exclude_pattern: /auth_views/
```
kennyadsl, line 8: I don't understand why we need this part: `exclude_pattern: /auth_views/`. There is no `auth_views` folder in the app directory, isn't it?

lib/solidus_starter_frontend/engine.rb
```diff
@@ -9,6 +9,10 @@ class Engine < Rails::Engine
 
     engine_name 'solidus_starter_frontend'
 
+    initializer 'solidus_starter_frontend', before: 'solidus_support_frontend_paths' do
```
kennyadsl, line 12: Do we need to add `solidus_support_frontend_paths` [here](https://github.com/solidusio/solidus_support/blob/09bfde4e67496167aca79cf23f05c7d8333031ac/lib/solidus_support/engine_extensions.rb)?

lib/generators/solidus_starter_frontend/solidus_starter_frontend_generator.rb
waiting-for-dev: It's because of this file:

https://github.com/nebulab/solidus_starter_frontend/blob/waiting-for-dev/fix_executable_installation/app/controllers/concerns/solidus_starter_frontend/auth_views.rb

We don't need it, thanks to your suggestion to add the views directly into the application. However, it crashes because of `Engine` not being within the namespace if we copy it.

lib/solidus_starter_frontend/engine.rb
waiting-for-dev: Yes, for not, it's added on my fork https://github.com/waiting-for-dev/solidus_support/blob/bcda06b7ff436018b5a3eb2588ca55a7128a7d5a/lib/solidus_support/engine_extensions.rb#L85. For now, you need to manually add it to the Gemfile if you want to test this branch.

waiting-for-dev:
@gsmendoza I think we can close this issue as you took another direction, right?

gsmendoza:
@waiting-for-dev Yes. Thanks for the work here! It helped me figure out how to move forward with the project :)

PR #166, opened 2021-07-09
Fix sandbox script

gsmendoza:
## Description

* Point solidus gems in sandbox Gemfile to those in
solidus_starter_frontend's Gemfile.

* Remove solidus_starter_frontend from sandbox's Gemfile.

* Fix: Install solidus using `solidus:install` instead of
`spree:install`.

* Fix: Pass `--auto-run-migrations` instead of `--auto-accept` to
`solidus:auth:install`.

* Run the solidus_starter_frontend installer directly from sandbox. Can
no longer call `rails generate solidus_starter_frontend:install` from
the sandbox because the extension has been removed from the sandbox's
Gemfile in this commit.

## How Has This Been Tested?

Manually ran `bin/sandbox`. 

Demo here: 

https://www.loom.com/share/c6022d758eed48beace61a5318cf996e.

## Types of changes

## Checklist:

### DIFF
--- a/bin/sandbox
+++ b/bin/sandbox
@@ -58,13 +58,19 @@ cat <<RUBY >> Gemfile
 # By default, the solidus gem also includes the standard frontend via
 # the solidus_frontend gem. To make this extension work, you need to
 # exclude it and manually include all the other Solidus components.
-gem 'solidus_core'
-gem 'solidus_api'
-gem 'solidus_backend'
-gem 'solidus_sample'
+
+solidus_repo = ENV.fetch('SOLIDUS_REPO', 'solidusio/solidus')
+solidus_branch = ENV.fetch('SOLIDUS_BRANCH', 'master')
+solidus_i18n_repo = ENV.fetch('SOLIDUS_I18N_REPO', 'solidusio/solidus_i18n')
+solidus_i18n_branch = ENV.fetch('SOLIDUS_I18N_BRANCH', 'master')
+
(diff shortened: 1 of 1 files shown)


bin/sandbox
```diff
@@ -93,16 +99,16 @@ unbundled bundle install --gemfile Gemfile
...
   --user_class=Spree::User \
   --enforce_available_locales=true \
   --with-authentication=false \
   --payment-method=none \
   $@
 
-unbundled bundle exec rails generate solidus:auth:install --auto-accept
-unbundled bundle exec rails generate ${extension_name}:install --auto-accept
+unbundled bundle exec rails generate solidus:auth:install --auto-run-migrations
+unbundled ../exe/${extension_name}
```
gsmendoza, line 111: I'm not sure if we still need to retain `unbundled` in this call. I tried running `../exe/solidus_starter_frontend`  directly from the sandbox directory and it seemed fine.

kennyadsl: Not sure either. I'd say to keep it until it gives us some problems, but up to you.

kennyadsl (approved):
👍 Thanks for the loom!

spaghetticode (approved):

gsmendoza:
Temporarily closing until I have a better understanding of https://github.com/nebulab/solidus_starter_frontend/pull/153 and https://github.com/nebulab/solidus_starter_frontend/pull/162.

PR #167, opened 2021-07-13
Discourage SolidusStarterFrontend from being run as an engine

gsmendoza:
## Dependencies

Depends on https://github.com/nebulab/solidus_starter_frontend/pull/166.

## Motivation and Context

From https://www.notion.so/nebulab/Starter-FE-7f00e5d23f5841c1817beebfc8859d83:

> We want the new starter frontend to be opinionated and provide a single option which is the "right" way for building your storefront, so we should only allow people to install the starter frontend through a Rails generator and remove (or at the very least severely restrict) the ability to run it as a Rails engine.

## Walkthrough and Demo

Demo: https://www.loom.com/share/a7be0f380b944cce9678796e6c88be12

## Types of changes

## Checklist:

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 require 'solidus_dev_support/rake_tasks'
+
+ENV['SOLIDUS_STARTER_FRONTEND_ALLOW_AS_ENGINE'] = 'true'
 SolidusDevSupport::RakeTasks.install
 
 task default: 'extension:specs'
--- a/docs/development.md
+++ b/docs/development.md
@@ -25,7 +25,17 @@ Simply add this require statement to your spec_helper:
 require 'solidus_starter_frontend/factories'
 ```
 
+### Running the extension in development
+
+In order to run the extension, you can either 1) run the sandbox script to
+generate the sandbox app or 2) add the extension as a gem in a Rails app.
+
 ### Running the sandbox
+
+The sandbox script uses the `solidus_starter_frontend:install` generator to
+install the frontend. This is useful when you want to ensure that the generator
+is working.
+
 To run this extension in a sandboxed Solidus application, you can run 
 `bin/sandbox`. The path for the sandbox app is `./sandbox` and `bin/rails` will 
 forward any Rails commands to `sandbox/bin/rails`.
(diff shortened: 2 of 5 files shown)


lib/solidus_starter_frontend/engine.rb
```diff
@@ -15,6 +15,8 @@ class Engine < Rails::Engine
     end
 
     config.to_prepare do
+      raise 'Error: SolidusStarterFrontend is not allowed to run as an engine' unless ENV['SOLIDUS_STARTER_FRONTEND_ALLOW_AS_ENGINE']
```
gsmendoza, line 18: I appreciate your suggestions on improving the copy of this error message.

spec/spec_helper.rb
```diff
@@ -2,6 +2,7 @@
 
 # Configure Rails Environment
 ENV['RAILS_ENV'] = 'test'
+ENV['SOLIDUS_STARTER_FRONTEND_ALLOW_AS_ENGINE'] = 'true'
```
gsmendoza, line 5: This is needed when running `bundle exec rspec`.

Rakefile
```diff
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 require 'solidus_dev_support/rake_tasks'
+
+ENV['SOLIDUS_STARTER_FRONTEND_ALLOW_AS_ENGINE'] = 'true'
```
gsmendoza, line 5: This is needed when running `bundle exec rake`.

docs/development.md
```diff
@@ -42,6 +52,82 @@ Use Ctrl-C to stop
...
+  --with-authentication=false \
+  --payment-method=none
+
+bundle exec rails generate solidus:auth:install --auto-run-migrations
+```
+
+#### 4) Install the frontend assets
+
+You will need to run `bundle exec rails g solidus_starter_frontend:install` to add the
+frontend assets to the existing vendored Solidus manifest files.
```
gsmendoza, line 123: I just realized now that instead of running this install command when running the gem as an engine, we could add the assets directly to `app/views/spree/layouts/spree_application.html.erb` and then strip them from the layout when running the generator. Let me know what you think!

docs/development.md
```diff
@@ -25,7 +25,17 @@ Simply add this require statement to your spec_helper:
 require 'solidus_starter_frontend/factories'
 ```
 
+### Running the extension in development
+
+In order to run the extension, you can either 1) run the sandbox script to
+generate the sandbox app or 2) add the extension as a gem in a Rails app.
+
 ### Running the sandbox
```
waiting-for-dev, line 33: Minor suggestion. Should this title be `####` (4 `#`) as it's within `Running the extension in development` (which is 3 `#`)?

docs/development.md
```diff
@@ -42,6 +52,82 @@ Use Ctrl-C to stop
 
 Default username and password for admin are: `admin@example.com` and `test123`.
 
+### Running the extension as an engine in a Rails app
```
waiting-for-dev, line 55: Same as before. Shouldn't it be `#` x 4? :slightly_smile:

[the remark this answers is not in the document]
waiting-for-dev: I think I don't fully understand why we need to provide two different development options. Shouldn't it be enough to have the sandbox load `solidus_starter_frontend` as a gem? If we point to the local gem (one directory up), changes will also be reflected. WDYT?

docs/development.md
gsmendoza: Yeah, I intentionally kept the two headings at a higher level for aesthetic purposes. I can move them lower.

docs/development.md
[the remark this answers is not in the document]
gsmendoza: @waiting-for-dev The two development options have different goals:

1. As a gem contributor, I want a easy way to test if a `solidus_starter_frontend` can generate a working frontend for a Rails app. 
2. As a gem contributor, I want a easy way to preview changes I make to the app code of `solidus_starter_frontend`.

Previously, the sandbox script was both loading the gem _and_ generating the frontend. In #166, I changed the sandbox script to fulfill _only_ goal 1, based on the reasoning that the gem will be used only as a generator in actual usage.

Goal 2, on the other hand, is more of a nice-to-have for the gem contributor, since it just makes it easier for the contributor to preview changes they make to the app code.

We could change the sandbox script to load the gem instead of using it to generate the frontend. Perhaps I might be misunderstanding what a sandbox app is supposed to do, based on conventions we follow with other Solidus extensions?

Your thoughts?

docs/development.md
waiting-for-dev: Maybe you're right, I didn't check the design :slightly_smiling_face:

gsmendoza: Nah, I think your suggestion works better. I lowered the headings.

docs/development.md
gsmendoza: I lowered the headings.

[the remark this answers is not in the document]
waiting-for-dev: Hmmm yeah, it makes sense. Ideally, I'd like to confirm that the `solidus_starter_frontend` executable can generate a working frontend in the integrated test suite. Still, if that's not possible, then I agree we need some way of at least doing it manually.

> Perhaps I might be misunderstanding what a sandbox app is supposed to do, based on conventions we follow with other Solidus extensions?

What do you mean by the conventions we follow with other extensions? I haven't worked with Solidus extensions, so I'm unfamiliar. Isn't the sandbox app there picking changes made to the extension code? At least, it's what I do on the main Solidus repo, so I had the idea it was one of its purposes.

Anyway, yeah, probably we need both strategies, but I definitely would automate both of them. Maybe we could even provide a flag to `bin/sandbox` so switch from one to the other, defaulting to the as-gem version as I think it'll be more common. However, if we take this direction, it'd be nice to be able to configure the application name so that generating it with another strategy doesn't forcefully means overwriting the previous one (we could even use always two different names, like `sandbox_as_gem` and `sandbox_as_exec` :upside_down_face: )

I'd like to ping @kennyadsl to hear what are his thoughts on this, too.

gsmendoza:
Temporarily closing until I have a better understanding of https://github.com/nebulab/solidus_starter_frontend/pull/153 and https://github.com/nebulab/solidus_starter_frontend/pull/162.

docs/development.md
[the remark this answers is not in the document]
gsmendoza: @waiting-for-dev I agree with defaulting to as-gem version since 1) this seems to be the more common way you'd want to use the sandbox, and 2) based on what you said, it's how the sandbox behaves in the main Solidus repo. 

I also like the idea of have two different scripts/apps for `sandbox_as_gem` and `sandbox_as_exec`.

PR #169, opened 2021-07-20
Fix and update sandbox script

gsmendoza:
Goal
----

I want the gem's sandbox script to generate the app with the `solidus_starter_frontend gem` as an engine
So that the app would pick up changes I make to the app code

Background
----------

Currently, the sandbox script both loads the gem and runs its generator. We want it to only the load the gem.

Previous implementations
------------------------

* [Document how to run the extension as an engine in a Rails app](https://github.com/nebulab/solidus_starter_frontend/pull/167/commits/a4600e2f3d37dec4eacdf24a36e68ebd124d31b8) - This commit documents how to install solidus_starter_frontend as an engine.
* Fix sandbox script #166 - This PR fixed the script to run `solidus_starter_frontend`  as a generator instead of as a gem.

Relevant links
--------------

https://github.com/nebulab/solidus_starter_frontend/pull/167#discussion_r669313064 - discussion on having two strategies for generating the sandbox app: one with the Starter FE running as an engine and other where Starter FE is generated into the app.

Implementation
--------------

* Point solidus gems in sandbox Gemfile to those in
solidus_starter_frontend's Gemfile.

* Fix: Install solidus using `solidus:install` instead of
`spree:install`.

* Fix: Pass `--auto-run-migrations` instead of `--auto-accept` to
`solidus:auth:install`.

* Remove `solidus_starter_frontend` call.

* Call `bundle exec rails g solidus_starter_frontend:install`.

## Demo

https://www.loom.com/share/1c7c4ff55baf4d8d8404a77f8520ca30

## Types of changes

## Checklist:

### DIFF
--- a/bin/sandbox
+++ b/bin/sandbox
@@ -58,12 +58,19 @@ cat <<RUBY >> Gemfile
 # By default, the solidus gem also includes the standard frontend via
 # the solidus_frontend gem. To make this extension work, you need to
 # exclude it and manually include all the other Solidus components.
-gem 'solidus_core'
-gem 'solidus_api'
-gem 'solidus_backend'
-gem 'solidus_sample'
+
+solidus_repo = ENV.fetch('SOLIDUS_REPO', 'solidusio/solidus')
+solidus_branch = ENV.fetch('SOLIDUS_BRANCH', 'master')
+solidus_i18n_repo = ENV.fetch('SOLIDUS_I18N_REPO', 'solidusio/solidus_i18n')
+solidus_i18n_branch = ENV.fetch('SOLIDUS_I18N_BRANCH', 'master')
+
+gem 'solidus_core', github: solidus_repo, branch: solidus_branch
+gem 'solidus_api', github: solidus_repo, branch: solidus_branch
+gem 'solidus_backend', github: solidus_repo, branch: solidus_branch
+gem 'solidus_sample', github: solidus_repo, branch: solidus_branch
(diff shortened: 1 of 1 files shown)


bin/sandbox
```diff
@@ -93,16 +100,16 @@ unbundled bundle install --gemfile Gemfile
...
   --user_class=Spree::User \
   --enforce_available_locales=true \
   --with-authentication=false \
   --payment-method=none \
   $@
 
-unbundled bundle exec rails generate solidus:auth:install --auto-accept
-unbundled bundle exec rails generate ${extension_name}:install --auto-accept
+unbundled bundle exec rails generate solidus:auth:install --auto-run-migrations
+unbundled bundle exec rails g solidus_starter_frontend:install
```
gsmendoza, line 112: An alternative to running the `solidus_starter_frontend:install` install command is to add the assets directly to `app/views/spree/layouts/spree_application.html.erb` and then strip them from the layout when running the generator. Let me know if you prefer this alternative.

kennyadsl: Not sure to get what you mean here, do you have an example?

gsmendoza: @kennyadsl I was easier to create a separate PR implementing that change :) Please see https://github.com/nebulab/solidus_starter_frontend/pull/170. Thanks!

kennyadsl: I'd just go with this PR's route for now.

kennyadsl (approved):

PR #170, opened 2021-07-21
Include assets in layout

gsmendoza:
Goal
----

I want the assets to be included in the layout

So that when I include the engine to a Solidus application, I no longer have to
run `bundle exec rails g solidus_starter_frontend:install` separately to add
the assets to the application.

Changes
-----------------------

* When I run `rails g solidus_starter_frontend:install`, the generator removes
  the asset tags from the layouts. The generator still adds those assets
  to the vendored all.* asset files.

## Dependencies

https://github.com/nebulab/solidus_starter_frontend/pull/169

## Demo

https://www.loom.com/share/393c57efb3d9411bae717405b6b6103c

## Types of changes

## Checklist:

### DIFF
--- a/app/views/spree/layouts/spree_application.html.erb
+++ b/app/views/spree/layouts/spree_application.html.erb
@@ -9,7 +9,10 @@
     <%= csrf_meta_tags %>
 
     <%= javascript_include_tag 'spree/frontend/all' %>
+    <%= javascript_include_tag 'spree/frontend/solidus_starter_frontend' %>
+
     <%= stylesheet_link_tag 'spree/frontend/all', media: 'screen' %>
+    <%= stylesheet_link_tag 'spree/frontend/solidus_starter_frontend', media: 'screen' %>
 
     <%= favicon_link_tag 'favicon.ico' %>
     <%= yield :head %>
--- a/bin/sandbox
+++ b/bin/sandbox
@@ -109,7 +109,6 @@ unbundled bundle exec rails generate solidus:install \
   $@
 
 unbundled bundle exec rails generate solidus:auth:install --auto-run-migrations
-unbundled bundle exec rails g solidus_starter_frontend:install
 
 echo
 echo "🚀 Sandbox app successfully created for $extension_name!"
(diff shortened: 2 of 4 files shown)


kennyadsl:
I have a couple of questions:

1. I'm not sure to understand why we are not using the existing `<%= javascript_include_tag 'spree/frontend/all' %>` assets. From what I recall, we can inject things into that manifest easily.

2. What about creating another generator for the sandbox only? Something like `solidus_starter_frontend:sandbox:install`?
This way we can use that generator to add the files to the application layout instead of adding + removing them later.

gsmendoza:
@kennyadsl

> I'm not sure to understand why we are not using the existing `<%= javascript_include_tag 'spree/frontend/all' %>` assets. From what I recall, we can inject things into that manifest easily.

I believe that is what the sandbox script is doing in PR #169: it uses `<%= javascript_include_tag 'spree/frontend/all' %>` in the layout and then calls `rails g solidus_starter_frontend:install` to inject `//= require spree/frontend/solidus_starter_frontend\n` to the `vendor/assets/javascripts/spree/frontend/all.js` manifest.

For me, that solution is okay. However, considering that the `sandbox` script is already adding the `solidus_starter_frontend` gem to the sandbox app's Gemfile, it seems nice if we can avoid calling that extra `rails g solidus_starter_frontend:install` command.

In any case, either solution (#169 or this PR) is okay with me.

> What about creating another generator for the sandbox only? Something like `solidus_starter_frontend:sandbox:install`? This way we can use that generator to add the files to the application layout instead of adding + removing them later.

Hmmm... I think that's not much different from #169, where we add `solidus_starter_frontend` as an engine to the sandbox app and then call the `rails g solidus_starter_frontend:install` generator to update the manifest.

To recap, I'm good with the solution in #169. If you think the #169 solution is clearer and more obvious, I'm good with closing this PR.

gsmendoza:
Per https://github.com/nebulab/solidus_starter_frontend/pull/169#discussion_r673846449, I'm closing this PR.

PR #171, opened 2021-07-22
Create alternate sandbox script which copies Starter FE into app

gsmendoza:
Goal
----

I want an alternate sandbox script which copies Starter FE into the sandbox
application

So that I have an easy way to confirm if the Starter FE can generate a working
frontend for a Rails app.

Previous implementations
------------------------

* Fix sandbox script https://github.com/nebulab/solidus_starter_frontend/pull/166.

Relevant links
--------------

https://github.com/nebulab/solidus_starter_frontend/pull/167#discussion_r669313064 - discussion on having two strategies for generating the sandbox app: one with the Starter FE running as an engine and other where Starter FE is generated into the app.

## Walkthrough

https://www.loom.com/share/85df7cc6cdae45669b36d57e5b8ed7eb

## Tests

* Create sandbox app with gem running as an engine - https://www.loom.com/share/6751affbc4f34350b98f1426cdf16f16
* Create sandbox app with storefront generated by the gem - https://www.loom.com/share/780222f03b4446ce86f0939e503c3439

## Types of changes

## Checklist:


kennyadsl:
I love those loom videos, thanks for that. They help a lot reviewing here. 

Just a couple of notes: 

1. I noticed that in the generated sandbox there is no login link. Do we expect that?
2. The term "generated" doesn't sound super clear to me, it can mean multiple things and I'm afraid that people will have hard time understanding its real meaning. What about something like just `sandbox` for the generated one and `sandbox --as-engine` for the engine?
3. Can we add some details in the README about how to use sandboxes?

gsmendoza:
Hi @kennyadsl! 

> I noticed that in the generated sandbox there is no login link. Do we expect that?

Hmm.. I'll investigate that.

> What about something like just sandbox for the generated one and sandbox --as-engine for the engine?

I would prefer if `bin/sandbox` creates an engine-based app since that is consistent with the `bin/sandbox` in https://github.com/solidusio/solidus_dev_support/blob/ead675245b0a0aee7b9216b5af9b7cd6d4285009/lib/solidus_dev_support/templates/extension/bin/sandbox.tt. Also, I think it would make sense to use the shorter "sandbox" name for the engine-based app, since I think it will be utilized more than the generated-storefront app. The engine-based app is more convenient in gem development since it automatically loads app changes.

The only other word I can think of other than "generated" is "installed". How about these alternatives?

* sandbox-generated-storefront
* sandbox-installed-storefront

If we can't come up with a better name, I'm okay going with your suggestion (sandbox and sandbox-as-engine).

> Can we add some details in the README about how to use sandboxes?

Okay, I'll update the https://github.com/nebulab/solidus_starter_frontend#development section.

kennyadsl:
> I would prefer if bin/sandbox creates an engine-based app since that is consistent with the bin/sandbox in https://github.com/solidusio/solidus_dev_support/blob/ead675245b0a0aee7b9216b5af9b7cd6d4285009/lib/solidus_dev_support/templates/extension/bin/sandbox.tt

Fair point! 

For the name, I'm still not convinced, let see if @aldesantis or @waiting-for-dev have some other ideas.

gsmendoza:
> I noticed that in the generated sandbox there is no login link. Do we expect that?

@kennyadsl Looking at the code, I suspect this is the auth-devise issue that https://github.com/nebulab/solidus_starter_frontend/pull/162 is supposed to fix. When I try going to the login URL, I get an "uninitialized constant Spree::UserSessionsController Did you mean? Spree::UserStockLocation" error.

Although that PR provides a working fix to the issue, I understand that the direction we have chosen to take is to install the solidus_auth_devise frontend code with a generator (https://github.com/solidusio/solidus_support/pull/66#issuecomment-864983185). 

I'll close this PR for now so that I can focus on the solidus_auth_devise frontend generator.

bin/sandbox
```diff
@@ -33,13 +33,28 @@ fi
...
+  sandbox_name='sandbox'
+  extension_gem_line="gem '$extension_name', path: '..'"
+else
+  sandbox_name='sandbox-generated'
+  extension_gem_line=""
+
+  if ! gem list --local | grep $extension_name; then
+    echo "ERROR: Please install $extension_name to generate the frontend of the sandbox."
+    exit 1
+  fi
```
gsmendoza, line 46: `solidus_starter_frontend` must be installed globally in order to create the sandbox-generated app.

README.md
```diff
@@ -79,8 +79,15 @@ with the next versions released since this project's gem will not be present in
...
+
 For information about contributing to this project please refer to this
-[document](docs/development.md).
+[document](docs/development.md). There you'll find information on tasks like:
+
+* Testing the extension
+* Running the sandbox
+* Updating the changelog
+* Releasing new versions
+* Docker development
```
gsmendoza, line 90: Provided more details on the Development documentation so that developers will know where to find more information on tasks such as how to run the sandbox. Addresses https://github.com/nebulab/solidus_starter_frontend/pull/171#issuecomment-884715051.

gsmendoza:
@aldesantis  @waiting-for-dev I am reopening this PR. I appreciate if you can continue the review.


### Known issues

At this point, the sandbox-generated app does not have working authentication yet. Thus, you will notice that the login link is not appearing on the app. See @kennyadsl's comment (https://github.com/nebulab/solidus_starter_frontend/pull/171#issuecomment-884715051).

I will be submitting https://github.com/nebulab/solidus_starter_frontend/pull/172 in order to address that issue.


### Updates since last review

* Require solidus_starter_frontend to be installed when running script to generate the frontend. https://github.com/nebulab/solidus_starter_frontend/pull/171#discussion_r678933726
* Added details in the README on where to find sandbox information. https://github.com/nebulab/solidus_starter_frontend/pull/171#discussion_r678935105


### Requested feedback

* What name can you suggest for the sandbox-generated directory? https://github.com/nebulab/solidus_starter_frontend/pull/171#issuecomment-884770954

bin/sandbox
waiting-for-dev: The problem I see is that the developer probably doesn't want to install the released version of `solidus_starter_frontend`. Instead, they want to check that their changes didn't break the installation process. They can indeed install the extension by `gem build + gem install /local/gem/file`, but I think we can automate it. See my comment below.

bin/sandbox
```diff
@@ -109,7 +124,12 @@ unbundled bundle exec rails generate solidus:install \
   $@
 
 unbundled bundle exec rails generate solidus:auth:install --auto-run-migrations
-unbundled bundle exec rails g solidus_starter_frontend:install
+
+if [[ -z $GENERATE_FRONTEND ]]; then
+  unbundled bundle exec rails g solidus_starter_frontend:install
+else
+  unbundled ../exe/${extension_name} --auto-accept
```
waiting-for-dev, line 131: Even if we're running the executable from our local copy, it'll run load `lib` files from the global `solidus_starter_frontend` gem, which could be the one released in rubygems.org.

However, we can execute ruby with the `-I` option to add a directory to the `$LOAD_PATH`. In addition, I don't think we need `unbundled` here, as we're not running `bundle`. So, I think that we could remove the checking about `solidus_starter_frontend` being installed and instead change this line to:

```suggestion
  ruby -I../lib ../exe/${extension_name} --auto-accept
```

gsmendoza: Thanks @waiting-for-dev ! I updated the `solidus_starter_frontend` call here and remove the error checking at https://github.com/nebulab/solidus_starter_frontend/pull/171#discussion_r678933726.

docs/development.md
```diff
@@ -42,6 +49,16 @@ Use Ctrl-C to stop
 
 Default username and password for admin are: `admin@example.com` and `test123`.
 
+#### App with storefront generated by the gem
+
+To create a sandboxed Solidus application where the storefront is generated with
+this gem, run `GENERATE_FRONTEND=1 bin/sandbox`. The path for the sandbox app
+is `./sandbox-generated`. Unlike the previous `./sandbox` app, you'll need to
+`cd` to the `./sandbox-generated` directory in order to run the application.
```
waiting-for-dev, line 57: Do you think we could use an env var to dispatch to `sandbox-generated` when given?

bin/sandbox
waiting-for-dev: Thanks, @gsmendoza! Awesome! :heart:

docs/development.md
gsmendoza: @waiting-for-dev You mean, something like `FORWARD=sandbox-generated bin/rails`? I think we can update `bin/rails` to support that:

```rb
#!/usr/bin/env ruby

if %w[g generate].include? ARGV.first
  exec "#{__dir__}/rails-engine", *ARGV
elsif ['sandbox', 'sandbox-generated'].include?(ENV['FORWARD'])
  exec "#{__dir__}/rails-#{ENV['FORWARD']}", *ARGV
elsif ENV['FORWARD'].present?
  raise "Unknown value for FORWARD environment variable '#{ENV['FORWARD']}'"
else
  exec "#{__dir__}/rails-sandbox", *ARGV
end
```

What do you think?

waiting-for-dev: Yeah, I mean doing something like that. However, we could lower the cognitive burden if we use the same environment variable to generate the sandbox app and forward it to it through `bin/rails`. What do you think if we use `GENERATED` for both `GENERATE_FRONTEND` and `FORWARD`? I really don't matter about the name that much as long as it's the same for both scenarios.

kennyadsl (approved):
👍 waiting for the answer to the last @waiting-for-dev comment before merging, thanks!

bin/rails-sandbox
```diff
@@ -1,12 +1,12 @@
 #!/usr/bin/env ruby
 
-app_root = 'sandbox'
+app_root = ENV['GENERATE_FRONTEND'] ? 'sandbox-generated' : 'sandbox'
```
gsmendoza, line 3: @waiting-for-dev Regarding https://github.com/nebulab/solidus_starter_frontend/pull/171#discussion_r684947097, I was able to simplify the `bin/rails` script and allow it to create the sandbox-generated app and dispatch to it immediately. Here is the usage for the `bin/rails` script:

* `bin/rails g` - Dispatches to gem Rails engine
* `bin/rails ARGV` - Creates sandbox if needed and dispatches ARGV to it.
* `GENERATE_FRONTEND=1 bin/rails ARGV` - Creates sandbox-generated if needed and dispatches ARGV to it.

docs/development.md
gsmendoza: @waiting-for-dev Please see https://github.com/nebulab/solidus_starter_frontend/pull/171#discussion_r685716524.

bin/rails-sandbox
waiting-for-dev: Amazing, @gsmendoza !! :clap: :clap: :clap:

waiting-for-dev (approved):

PR #172, opened 2021-07-29
Copy routes/controllers/views/tests from auth_devise

gsmendoza:
## Goal

Allow devs to directly touch the authentication component of the storefront in their application.

## Dependencies

Indirectly depends on https://github.com/nebulab/solidus_starter_frontend/pull/171. That PR adds a script for creating a sandbox with generated frontend, which is useful for testing this PR.

## Walkthrough

https://www.loom.com/share/e58d10c8cc6b402e8408b1b3e70b375b

## Demo

* Browser demo - https://www.loom.com/share/90dd2709e34c4cd9a7f4598fa352e240
* Run system specs on generated sandbox - https://www.loom.com/share/3d8ab1ef74a5490eae16e4d4130a4eff

## Changes

* Copy storefront routes/controllers/views/tests from solidus_auth_devise into solidus_starter_frontend .
* Minimize solidus_starter_frontend's dependency on solidus_auth_devise.
* Ensure solidus_auth_devise loads frontend components only once, by ensuring that `frontend_available?` does not get set to true.

## Alternatives considered

Chose to add the `solidus_auth_devise` frontend to `solidus_starter_frontend` instead of adding the generator to auth devise primarily because I found this to be the simpler approach. By having the checkout and the authentication code in the same project, it's easier to verify if they are working together.

The downside with this approach is that frontend code is currently duplicated here and in `solidus_auth_devise`. However, since we're planning to replace the old `solidus_frontend` with this gem, the code in frontend code in `solidus_auth_devise` will eventually get removed.

## Previous attempts

* #153

## Relevant PRs

* #162

## Scope

These tasks are not covered in this PR and will be implemented in future PRs.

* Investigate `SolidusStarterFrontend::Views::OverrideGenerator` and update if necessary.
* Copy `solidus_auth_devise` unit tests to `solidus_starter_frontend`. For this PR, I found that the `solidus_auth_devise` feature specs were sufficient in verifying that the code migration has been set up properly.
* Add an option to exclude `solidus_auth_devise` and its frontend components from the generated storefront.

## Types of changes

- [N/A] New feature (non-breaking change which adds functionality)

## Checklist:


gsmendoza:
Regarding the caching test failures at https://www.loom.com/share/3d8ab1ef74a5490eae16e4d4130a4eff, I found that this is due to `config.cache_store` being set to `:null_store` in the `sandbox-generated` app. Disabling `config.cache_store = :null_store` in `config/environments/test.rb` fixes the specs.

spec/system/checkout_spec.rb
```diff
@@ -611,6 +617,7 @@
 
       it "displays the entered state name without evaluating" do
```
gsmendoza, line 618: This example seems fails frequently. See for example, https://app.circleci.com/pipelines/github/nebulab/solidus_starter_frontend/879/workflows/778f9ce5-62d3-43b4-af6b-6e282b275815/jobs/2243/tests. In the test run I did in the video below, I got it to fail 40 out of 100 times.

I discovered that the example fails frequently under the following conditions:

1. When signing in as a guest during checkout.
2. When using the apparition JS driver.

However, under ANY of these conditions, the example succeeds most of the time:

1. On the current master branch, where we don't have authentication yet.
2. When signing in as a user during checkout.
3. When using a different JS driver, like selenium_chrome_headless .

Please see the video below for more details. 

https://www.loom.com/share/2a6db47101f24fcd8206361f8300d7f9

For the meantime, I'll add a commit to apply the workaround I did in the video, i.e. sign in as user instead of gueset.

gsmendoza: Pushed commit https://github.com/nebulab/solidus_starter_frontend/pull/172/commits/effe34d81bd7ee07a9dc34ce5e990e9331667727.

app/controllers/spree/checkout_controller.rb
```diff
@@ -16,10 +16,15 @@ class CheckoutController < Spree::StoreController
     before_action :ensure_valid_state
 
     before_action :associate_user
+    before_action :check_registration, except: [:registration, :update_registration]
     before_action :check_authorization
 
     before_action :setup_for_current_state, only: [:edit, :update]
 
+    # This action builds some associations on the order, ex. addresses, which we
+    # don't to build or save here.
```
waiting-for-dev, line 25:
```suggestion
    # don't need to build or save here.
```

app/controllers/spree/checkout_controller.rb
```diff
@@ -47,6 +52,20 @@ def update
       end
     end
 
+    def registration
+      @user = Spree::User.new
+    end
+
+    def update_registration
+      if params[:order][:email] =~ Devise.email_regexp && current_order.update(email: params[:order][:email])
```
waiting-for-dev, line 60: Just double-checking, do we always have a `current_order` present?

waiting-for-dev: Forget it; I guess you just copied it here, so it should already have been checked.

waiting-for-dev (approved):
Amazing job @gsmendoza!

kennyadsl (approved):
Feel free to merge when ready. We can tackle the flaky spec issue in another PR, thanks!

PR #174, opened 2021-08-02
Extract VendorAssetsGenerator

gsmendoza:
Goal
----

I want to create a separate
`SolidusStarterFrontend::VendorAssetsGenerator`

So that the purposes and differences of the `SolidusStarterFrontend`
generators would be clearer:

* `SolidusStarterFrontend::Generators::InstallGenerator`
  - Used for setting up the extension in CI (see
    `solidus/core/lib/spree/testing_support/common_rake.rb`)
* `SolidusStarterFrontend::VendorAssetsGenerator`
  - Used for adding the JS and CSS files to the vendored all.*manifests.
* `SolidusStarterFrontendGenerator`
  - Called by users of the gem to install the starter frontend in their
    apps.

## Dependencies

Depends on https://github.com/nebulab/solidus_starter_frontend/pull/171

## Walkthrough

https://www.loom.com/share/5fecced6c0c9466da0157f8bb50cf905

## Demo

https://www.loom.com/share/a5ede129d3c14c938af543c0f4b04f86

## Types of changes

- [N/A] New feature (non-breaking change which adds functionality)

## Checklist:

- [N/A] My code follows the code style of this project.

### DIFF
--- a/lib/generators/solidus_starter_frontend/solidus_starter_frontend_generator.rb
+++ b/lib/generators/solidus_starter_frontend/solidus_starter_frontend_generator.rb
@@ -32,6 +32,17 @@ class SolidusStarterFrontendGenerator < Rails::Generators::Base
     inject_into_file 'config/initializers/spree.rb', "require_relative Rails.root.join('lib/solidus_starter_frontend/config')\n", before: /Spree.config do/, verbose: true
     gsub_file 'app/assets/stylesheets/application.css', '*= require_tree', '* OFF require_tree'
 
+    # We can't use Rails' `generate` method here to call
+    # `solidus_starter_frontend:vendor_assets`. When the
+    # solidus_starter_frontend generator is used as a standalone program
+    # (instead of added to an app's Gemfile), `generate` won't be able to pick
+    # up the other generators that the gem provides.
+    #
+    # We're able to use Thor's `invoke` action here instead of `generate`.
+    # However, `invoke` only works once per generator you want to call. Please
+    # see https://stackoverflow.com/a/12029262/65925 for more details.
+    #
+    # See also https://github.com/nebulab/solidus_starter_frontend/pull/174#discussion_r685626737.
     invoke 'solidus_starter_frontend:vendor_assets'
   end
 end
--- a/lib/generators/solidus_starter_frontend/solidus_starter_frontend_generator.rb
+++ b/lib/generators/solidus_starter_frontend/solidus_starter_frontend_generator.rb
@@ -29,9 +29,9 @@ class SolidusStarterFrontendGenerator < Rails::Generators::Base
 
     # Text updates
     append_file 'config/initializers/assets.rb', "Rails.application.config.assets.precompile += ['solidus_starter_frontend_manifest.js']"
-    append_file 'vendor/assets/javascripts/spree/frontend/all.js', "//= require spree/frontend/solidus_starter_frontend\n"
-    inject_into_file 'vendor/assets/stylesheets/spree/frontend/all.css', " *= require spree/frontend/solidus_starter_frontend\n", before: %r{\*/}, verbose: true
     inject_into_file 'config/initializers/spree.rb', "require_relative Rails.root.join('lib/solidus_starter_frontend/config')\n", before: /Spree.config do/, verbose: true
     gsub_file 'app/assets/stylesheets/application.css', '*= require_tree', '* OFF require_tree'
+
+    invoke 'solidus_starter_frontend:vendor_assets'
   end
 end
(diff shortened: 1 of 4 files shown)


waiting-for-dev:
I like it, better structure and modularization! :heart: 

Do you think we should also update `solidus_starter_frontend` generator to use the new `VendorAssetsGenerator`?

https://github.com/nebulab/solidus_starter_frontend/blob/8dd502ab5760d404324573be8220bb11350546cd/lib/generators/solidus_starter_frontend/solidus_starter_frontend_generator.rb#L32-L33

lib/generators/solidus_starter_frontend/solidus_starter_frontend_generator.rb
```diff
@@ -29,9 +29,9 @@ def install
 
     # Text updates
     append_file 'config/initializers/assets.rb', "Rails.application.config.assets.precompile += ['solidus_starter_frontend_manifest.js']"
-    append_file 'vendor/assets/javascripts/spree/frontend/all.js', "//= require spree/frontend/solidus_starter_frontend\n"
-    inject_into_file 'vendor/assets/stylesheets/spree/frontend/all.css', " *= require spree/frontend/solidus_starter_frontend\n", before: %r{\*/}, verbose: true
     inject_into_file 'config/initializers/spree.rb', "require_relative Rails.root.join('lib/solidus_starter_frontend/config')\n", before: /Spree.config do/, verbose: true
     gsub_file 'app/assets/stylesheets/application.css', '*= require_tree', '* OFF require_tree'
+
+    invoke 'solidus_starter_frontend:vendor_assets'
```
gsmendoza, line 35: @waiting-for-dev Regarding https://github.com/nebulab/solidus_starter_frontend/pull/174#issuecomment-894131810, I was able to reuse the `vendor_assets` generator here using Thor's `invoke` method. I found out about `invoke` from https://stackoverflow.com/a/12029262/65925. I initially tried Rail's [`generate` method](https://guides.rubyonrails.org/generators.html#generate) but the `solidus_starter_frontend` wasn't able to find the generator. Please see the video below:

https://www.loom.com/share/c27a99356ccf41c4a99f092f0704d099

waiting-for-dev: Oh, I see. `generate` seems unavailable because `solidus_starter_frontend` is not in the `Gemfile`, so the Rails context can't pick up the generators it provides.

As long as `VendorAssetsGenerator` remains simple, it'll work. I took a look at the post on SO you shared, and another caveat is that `invoke` [wouldn't dispatch arguments if needed](https://stackoverflow.com/questions/5890650/how-to-get-generators-call-other-generators-in-rails-3#comment34496472_12029262).

What do you think if we add a comment with all of that so that we're not caught in the future?

waiting-for-dev (approved):

lib/generators/solidus_starter_frontend/solidus_starter_frontend_generator.rb
gsmendoza: Sounds good. I'll add comments.

lib/generators/solidus_starter_frontend/solidus_starter_frontend_generator.rb
```diff
@@ -29,9 +29,20 @@ def install
...
+    # `solidus_starter_frontend:vendor_assets`. When the
+    # solidus_starter_frontend generator is used as a standalone program
+    # (instead of added to an app's Gemfile), `generate` won't be able to pick
+    # up the other generators that the gem provides.
+    #
+    # We're able to use Thor's `invoke` action here instead of `generate`.
+    # However, `invoke` only works once per generator you want to call. Please
+    # see https://stackoverflow.com/a/12029262/65925 for more details.
+    #
+    # See also https://github.com/nebulab/solidus_starter_frontend/pull/174#discussion_r685626737.
```
gsmendoza, line 45: @waiting-for-dev Added comment to address https://github.com/nebulab/solidus_starter_frontend/pull/174/files#r685672184.

waiting-for-dev: LGTM. Thanks!!

kennyadsl:
Can we take a step back to understand better why this is needed? If we are using Solidus copying all files in the host application, why are we still using the css/js provided by the gem? Are those files copied as well? If yes, maybe we could just copy them in a directory that doesn't require adding anything to the manifest?

gsmendoza:
> If yes, maybe we could just copy them in a directory that doesn't require adding anything to the manifest?

@kennyadsl That makes sense. Let me try it out. I'll close this PR for the meantime.

PR #175, opened 2021-08-12
Gsmendoza/eng 204 update solidusstarterfrontendgenerator

gsmendoza:
Goal
----

As a solidus_starter_frontend contributor
I want SolidusStarterFrontendGenerator to copy the test suite to the storefront app
So that I can the run the test suite to verify that the generated frontend is working

Considerations
--------------

We still want the specs to run on the gem. That is, we shouldn't be forced to
generate the app in order to run the test suite there.

Implications
------------

This would mean that SolidusStarterFrontendGenerator would also generate a
starter test suite for the storefronts it would generate.

*Pros*

* Users of SolidusStarterFrontendGenerator can lean on the test suite as they
  change their storefronts.

*Cons*

* We would be including the whole test suite, which may include
  * Low-quality tests
  * Tests that would make more sense in the context of the gem than in the
    context of a generated storefront.
* We could be bloating the user's test suite.

Thoughts
--------

Despite the cons, I think it would be difficult to curate a test suite
specifically for the generated storefront.

Furthermore, considering that we're already allowing the frontend code to be
consumed directly by users, I think that it's okay to afford the test suite the
same leniency we are allowing to the frontend code. As a plus, any improvements
we make later on to the frontend code and its test suite will benefit both the
gem and its consumers using its generated frontends.

Scope
-----

In a succeeding ticket, we can update the CI to run the specs in the generated app.

## Walkthrough

TODO

## Demos

TODO

## Types of changes

## Checklist:


spec/solidus_starter_frontend_helper.rb
```diff
@@ -0,0 +1,54 @@
...
+
+RSpec.configure do |config|
+  config.disable_monkey_patching!
+  config.infer_spec_type_from_file_location!
+  config.use_transactional_fixtures = false
+
+  # We currently have examples wherein we mock or stub method that do not exist on
+  # the real objects.
+  config.mock_with :rspec do |mocks|
+    mocks.verify_partial_doubles = false
```
gsmendoza, line 26: This would override a `mocks.verify_partial_doubles = true` config on the user's Rails app `spec_helper.rb` if one is set. We need to set this this to false in order to get the following tests to pass:

* `./spec/controllers/spree/base_controller_spec.rb`
* `./spec/controllers/spree/checkout_controller_spec.rb`
* `./spec/controllers/spree/products_controller_spec.rb`
* `./spec/requests/spree/orders_ability_spec.rb`

spec/solidus_starter_frontend_helper.rb
```diff
@@ -0,0 +1,54 @@
...
+    sign_in(user)
+  end
+
+  config.before(:each, with_guest_session: true) do
+    allow_any_instance_of(ActionDispatch::Cookies::CookieJar).to receive(:signed) { { guest_token: order.guest_token } }
+  end
+
+  config.around(:each, caching: true) do |example|
+    original_cache_store = ActionController::Base.cache_store
+    ActionController::Base.cache_store = ActiveSupport::Cache.lookup_store(:memory_store)
```
gsmendoza, line 48: By default, the cache store of test environment in a regular Rails app is `:null_store`. This config is needed to get the following specs to pass:

* `spec/system/caching/products_spec.rb`
* `spec/system/caching/taxons_spec.rb`

spec/solidus_starter_frontend_helper.rb
kennyadsl: Good to know. Yes, I also don't like this, what about trying to adjust specs that require the verification?

kennyadsl (approved):

spec/solidus_starter_frontend_helper.rb
gsmendoza: @kennyadsl Hmm... Let me see if I can adjust the specs before I merge this PR. TODO.

gsmendoza: @kennyadsl Turns out that most of the stubs were no longer needed :)

lib/generators/solidus_starter_frontend/rspec_generator.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+module SolidusStarterFrontend
+  class RspecGenerator < Rails::Generators::Base
+    source_root File.expand_path('../../..', __dir__)
+
+    def install
+      # solidus_dev_support 2.5.0 depends on this puma version.
+      gsub_file 'Gemfile', /gem ['"]puma['"].*/, "# gem 'puma' # solidus_dev_support 2.5.0 depends on puma ~> 4.3. See puma declaration below."
```
waiting-for-dev, line 9: It's quite unfortunate that we need to do that, even more considering that `solidus_dev_support` is not a dependency on production. I bet that we could update the `puma` version on `solidus_dev_support` without issues. We could do that as a follow-up and then open another PR updating from `solidus_starter_frontend`. Or, at least, we should open an issue on `solidus_dev_support` so that we don't lose track of it.

spec/controllers/spree/base_controller_spec.rb
```diff
@@ -9,7 +9,7 @@ def index; authorize!(:read, :something); end
     end
 
     before do
-      stub_spree_preferences(Spree::Config, redirect_back_on_unauthorized: true)
+      allow(Spree::Auth::Engine).to receive(:redirect_back_on_unauthorized?).and_return(true)
```
waiting-for-dev, line 12: Why is it needed?

spec/rails_helper.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+# This is a placeholder for `spec/rails_helper.rb`. We're assuming that the
+# Solidus app that will use solidus_starter_frontend will have this helper in its
+# spec directory.
+
+require 'solidus_starter_frontend_helper'
```
waiting-for-dev, line 7: Nice trick, it makes things less messy :clap:

spec/spec_helper.rb
```diff
@@ -39,23 +26,59 @@
...
-
-  if Spree.solidus_gem_version < Gem::Version.new('2.11')
-    config.extend Spree::TestingSupport::AuthorizationHelpers::Request, type: :system
-  else
-    config.include Spree::TestingSupport::Translations
+  config.mock_with :rspec do |mocks|
+    # Prevents you from mocking or stubbing a method that does not exist on
+    # a real object. This is generally recommended, and will default to
+    # `true` in RSpec 4.
+    mocks.verify_partial_doubles = true
```
waiting-for-dev, line 33: Good job updating the test suite to support the default and safe behavior!

waiting-for-dev (commented):
Superb work @gsmendoza, keeping track of all the nitpicky details :heart: 

As you said in the issue description, moving all the tests to the host application has several drawbacks, but it's consistent with the direction we're taken. I wonder if we could add an environment variable `SKIP_SPECS` so that users can choose not to copy them, though. WDYT?

lib/generators/solidus_starter_frontend/rspec_generator.rb
gsmendoza: @waiting-for-dev Okay, I'll create the tickets. TODO.

spec/controllers/spree/base_controller_spec.rb
gsmendoza: We need the stub (and the whole spec) to test `config/initializers/solidus_auth_devise_unauthorized_redirect.rb`.

gsmendoza:
> I wonder if we could add an environment variable SKIP_SPECS so that users can choose not to copy them, though. WDYT?

Yeah, that sounds good. Another important reason for adding this option is that the user might already have or might prefer TestUnit. I'll spike this first. TODO.

spec/spec_helper.rb
```diff
@@ -6,25 +6,12 @@
...
-require 'spree/testing_support/translations' unless Spree.solidus_gem_version < Gem::Version.new('2.11')
-
-# Requires supporting ruby files with custom matchers and macros, etc,
-# in spec/support/ and its subdirectories.
-Dir["#{__dir__}/support/**/*.rb"].sort.each { |f| require f }
-
-SolidusDevSupport::TestingSupport::Factories.load_for(SolidusStarterFrontend::Engine)
+require 'capybara'
 
 Capybara.register_driver :apparition_docker_friendly do |app|
```
waiting-for-dev, line 16: Could we move the registration of the driver to `spec/support/solidus_starter_frontend/capybara.rb`? We need to make the tests pass when executed through Docker from within the `sandbox` application. I know this is something related to the development of the gem in contrast to the host application, but the same can be said of the other registered driver (`:apparition`), so I guess we don't have a better option :man_shrugging:

gsmendoza: Okay, I'll move the driver to `spec/support/solidus_starter_frontend/capybara.rb`. Todo.

gsmendoza:
@waiting-for-dev I've updated the PR. Please see the following commits:

* Add option to skip specs when installing Solidus Starter Frontend
* Move apparition_docker_friendly driver to capybara helper
* Fix: confirmation spec fails in Docker container

Here's a walkthrough and demo of the "skip-specs" option added to SolidusStarterFrontendGenerator:

https://www.loom.com/share/a016691596384676b779a2f68dbd7aa9

spaghetticode (approved):
@gsmendoza amazing work, thank you! 🤩

spec/spec_helper.rb
```diff
@@ -6,56 +6,67 @@
...
-    headless: true,
-    browser_options: [
-      :no_sandbox,
-      :disable_gpu,
-      { disable_features: 'VizDisplayCompositor' }
-    ]
-  }
-  Capybara::Apparition::Driver.new(app, opts)
-end
+require 'capybara'
```
waiting-for-dev, line 14: Now that we have moved the registration of the driver, do we need this line?

waiting-for-dev (approved):

waiting-for-dev:
> @waiting-for-dev I've updated the PR. Please see the following commits:

Thank you @gsmendoza!!!

PR #177, opened 2021-08-17
Invert dependency of SolidusStarterFrontend and Solidus assets

gsmendoza:
## Dependencies

Depends on #175.

Goal
----

As a SolidusStarterFrontend contributor

I want to invert the dependencies of SolidusStarterFrontend and Solidus
assets such that the SolidusStarterFrontend manifests depend on the
Solidus assets (and not the other way around)

So that we can simplify the frontends generated by the gem

Background
----------

From
https://github.com/nebulab/solidus_starter_frontend/pull/174#issuecomment-896682012:

> Can we take a step back to understand better why this is needed? If we
are using Solidus copying all files in the host application, why are we
still using the css/js provided by the gem? Are those files copied as
well? If yes, maybe we could just copy them in a directory that doesn't
require adding anything to the manifest?

## Walkthrough

https://www.loom.com/share/dff302e4a8fc4e3b9de2dafd99e10c3f

## Tests

https://www.loom.com/share/914adea46a4b4c65af5dce08c7f518ec

## Types of changes

## Checklist:


waiting-for-dev (approved):
Noice!

spaghetticode (approved):
LGTM, thanks @gsmendoza 👍

PR #182, opened 2021-09-06
Add option to exclude authentication component when running generator

gsmendoza:
Dependencies
----

Depends on https://github.com/nebulab/solidus_starter_frontend/pull/183. This is currently on Draft mode until that PR is merged.  

Goal
----

As a solidus_starter_frontend user

I want solidus_starter_frontend to have an option to run without generating the
authentication component

So that I can choose to exclude the authentication component if I don't need it
or if I want to implement it in another way.

Done
----

* Update CI - Add run-specs-with-postgres-for-generators and
  run-specs-with-postgres-skipping-authentication jobs.

* bin/sandbox - Respond to SANDBOX_SKIP_AUTHENTICATION environment variable.

* SolidusStarterFrontendGenerator and RspecGenerator - Add
  `--skip-authentication` options.

* Introduce EnableCode, DisableCode, and RemoveMarkers services.

* Identify and mark authentication-related and non-authentication-related files
  and code.

Implementation trade-offs
-------------------------

* Chose to keep marker detection strict in order to keep their corresponding
  regex matchers simple. For example, markers are assumed to have no trailing
  whitespace.

* Chose to surround markers with blank lines in order to improve readability.
  This results in some duplicate blank lines after the generators are ran.
  Fixing these blank lines would result in more complicated regex matchers. As
  such, I believe more sophisticated linters such as RuboCop would be better
  suited for fixing these duplicate blank lines.

Known limitations
-----------------

* The `EnableCode` service can only enable uncomment Ruby (`*.rb`) files. It
  doesn't handle ERB files yet, since we currently don't have ERB code that
  require uncommenting.

## Walkthrough

https://www.loom.com/share/3c528a13c6cb485f9f6d2a150f53a8ad

## Manual tests

https://www.loom.com/share/e5dc1966bd6d4868ac733385332e46fc

## Types of changes

- N/A - Bug fix (non-breaking change which fixes an issue)
- N/A - Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

### DIFF
--- a/bin/sandbox
+++ b/bin/sandbox
@@ -50,6 +50,16 @@ else
   rspec_rails_gem_line="gem 'rspec-rails', groups: [:development, :test]"
 fi
 
+if [[ "${SANDBOX_SKIP_AUTHENTICATION}" == '1' ]]; then
+  generator_skip_authentication_argument='--skip-authentication'
+  solidus_auth_devise_gem_line=''
+  solidus_install_user_class_argument=""
+else
+  generator_skip_authentication_argument=''
+  solidus_auth_devise_gem_line="gem 'solidus_auth_devise'"
+  solidus_install_user_class_argument="--user_class=Spree::User"
+fi
+
 rm -rf "${sandbox_path}"
 bundle exec rails new "${sandbox_name}" --database="${railsdb}" \
   --skip-bundle \
@@ -84,7 +94,7 @@ gem 'solidus_i18n', github: solidus_i18n_repo, branch: solidus_i18n_branch
 
 gem 'rails-i18n'
 ${extension_gem_line}
-gem 'solidus_auth_devise'
+${solidus_auth_devise_gem_line}
 
 ${rspec_rails_gem_line}
 RUBY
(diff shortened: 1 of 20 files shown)


bin/sandbox
```diff
@@ -45,8 +45,20 @@ sandbox_path="./${sandbox_name}"
 
 if [[ -n $SKIP_SPECS ]]; then
   GENERATOR_SKIP_SPECS_ARGUMENT='--skip-specs'
+  rspec_rails_gem_line=''
 else
   GENERATOR_SKIP_SPECS_ARGUMENT=''
+  rspec_rails_gem_line="gem 'rspec-rails', groups: [:development, :test]"
+fi
+
+if [[ $SANDBOX_SKIP_AUTHENTICATION == '1' ]]; then
```
gsmendoza, line 54: Instead of checking if it's an empty string like the other conditions  in this script, this condition checks against "1" because the `skip_authentication` parameter of the `test-branch` command in `.circleci/config.yml` sets the value to either "1" or "0". Thus, its value wouldn't be empty in the CI job.

lib/generators/solidus_starter_frontend/disable_code.rb
```diff
@@ -0,0 +1,48 @@
...
+    private
+
+    def delete_lines
+      marker_namespace.line_markers.each do |line_marker|
+        generator.gsub_file path, /.*#{line_marker}\n?/, ''
+      end
+    end
+
+    def delete_blocks
+      non_greedy_content_regex_string = "(?:.*?)"
```
gsmendoza, line 29: Using non-capturing groups as recommended by https://github.com/rubocop/ruby-style-guide#non-capturing-regexp: 

image

spec/generators/solidus_starter_frontend/disable_code_spec.rb
```diff
@@ -0,0 +1,221 @@
...
+
+module SolidusStarterFrontend
+  RSpec.describe DisableCode do
+    describe '#call' do
+      let(:generator_class) do
+        Class.new(Rails::Generators::Base) do
+        end
+      end
+
+      let(:generator) { generator_class.new([], quiet: true) }
```
gsmendoza, line 15: Figured out how to instantiate a generator object mostly by trial and error :p

gsmendoza:
Will try to implement this by separating the authentication files from the base installation. Closing this for now.

PR #184, opened 2021-09-16
Add option to exclude authentication component when running generator

gsmendoza:
## Description

As a solidus_starter_frontend user
I want solidus_starter_frontend to have an option to run without generating the authentication component
So that I can choose to exclude the authentication component if I don't need it or if I want to implement it in another way.

As a SolidusStarterFrontend contributor
I would like to remove the engine functionality of the gem
So that the gem would be simpler to maintain
And we can write files with conditional authentication code using Thor templates.

## Implementation

* This PR uses #182 as a starting point, since that PR has already identified the areas of the frontend code that have conditional authentication code.

## Scope

I'll tackle converting the gem to a Rails template in a separate PR following this.

## Walkthrough

https://www.loom.com/share/14e90808b0b1467c90873075e72ef51a

## Tests/Demos

* Run `solidus_starter_frontend` with authentication - https://www.loom.com/share/607e53afcc49457dba8f64a0dcb59668
* Run `solidus_starter_frontend` without authentication - https://www.loom.com/share/0f6d949288884560827101dc698486af

## Previous implementations

https://github.com/nebulab/solidus_starter_frontend/pull/182

## Types of changes

- N/A - Bug fix (non-breaking change which fixes an issue)

## Checklist:


Gemfile
```diff
@@ -1,29 +1,12 @@
...
-gem 'solidus_core', github: solidus_repo, branch: solidus_branch
-gem 'solidus_api', github: solidus_repo, branch: solidus_branch
-gem 'solidus_backend', github: solidus_repo, branch: solidus_branch
-gem 'solidus_sample', github: solidus_repo, branch: solidus_branch
-gem 'solidus_i18n', github: solidus_i18n_repo, branch: solidus_i18n_branch
 
 # Needed to help Bundler figure out how to resolve dependencies,
 # otherwise it takes forever to resolve them.
 # See https://github.com/bundler/bundler/issues/6677
 gem 'rails', '>0.a'
```
gsmendoza, line 8: `rails` is still needed by `bin/sandbox` to run the generator.

bin/solidus_compare
```diff
@@ -96,13 +96,14 @@ def load_config
   @config['source_name'] ||= 'solidus'
   @config['source_branch'] ||= 'master'
   @config['source_base_path'] ||= 'frontend/'
+  @config['local_base_path'] ||= 'templates/'
```
gsmendoza, line 99: Need to specify that the templates are in the `templates/` directory.

docs/development.md
```diff
@@ -2,40 +2,13 @@
 This document aims to give some extra information for developers that are going 
 to contribute to our `solidus_starter_frontend` component.
 
-### Testing the extension
```
gsmendoza, line 5: Removed obsolete development documentation.

templates/app/controllers/spree/checkout_controller.rb.tt
```diff
@@ -16,14 +16,20 @@ class CheckoutController < Spree::StoreController
     before_action :ensure_valid_state
 
     before_action :associate_user
+
+    <%- if include_authentication? -%>
```
gsmendoza, line 20: `include_authentication?` is a private method of SolidusStarterFrontendGenerator. When the Thor templates are processed by the SolidusStarterFrontendGenerator, the object that receives the messages from the ERB code is the generator instance.

lib/generators/solidus_starter_frontend/solidus_starter_frontend_generator.rb
```diff
@@ -2,27 +2,49 @@
 
 require 'bundler'
 
+# `rspec:install` requires these libraries to run.
+require 'stringio'
+require 'tmpdir'
```
gsmendoza, line 7: As mentioned by the comment, these libraries are required to invoke `rspec:install`. I haven't figure out why this has to be done, but we might eventually remove these requires once we convert the gem into a rails template.

gsmendoza:
@kennyadsl I've added some changes to restore and improve the CircleCI gem caching. Please see https://www.loom.com/share/367b91970c6c4319a3503daf9a910918 for the walkthrough. Thanks!

.circleci/config.yml
```diff
@@ -28,14 +31,29 @@ commands:
     steps:
       - run:
           name: 'Solidus <<parameters.branch>>: Generate Gemfile.lock'
-          command: bundle lock
+          command: |
+            rm Gemfile
+
+            cat \
```
kennyadsl, line 37: What about adding those things in the Gemfile directly. That Gemfile was used to run bundle locally and consider the dependencies specified in the gemspec. But now that we are going with using this project as a generator only I think we don't need that content anymore.

Gemfile
kennyadsl: I think we can remove the `'>0.a'` here. It was a trick needed for a scenario that is already solved.

gsmendoza:
@kennyadsl I'll be closing this PR. I have applied the feedback you gave to #185. It's a bit hard to revert changes here since they will break #185.

PR #193, opened 2021-10-27
Refactor page - checkout/edit

gsmendoza:
## Goal

As a solidus starter frontend user
I want the pages to be refactored 
So that they would be easier to extend and update

## How Has This Been Tested?

Manually tested.

image

## Types of changes

- N/A: Bug fix (non-breaking change which fixes an issue)
- N/A: New feature (non-breaking change which adds functionality)

## Checklist:
- [x ] My code follows the code style of this project.


templates/app/views/spree/checkout/_checkout_step.html.erb
```diff
@@ -0,0 +1,20 @@
+<%
+  form_url = update_checkout_path(@order.state)
```
kennyadsl, line 2: Why these variables? Can't we just have them in the code below, I can't see any advantage of having them here but maybe I'm missing something.

gsmendoza: Sure, I can inline them. TODO.

kennyadsl (approved):

PR #199, opened 2021-11-04
Move user error messages inside auth container

gsmendoza:
## Bug description

Error explanation in user forms are not aligned with the form.

### http://localhost:3000/signup

image

### http://localhost:3000/password/change

image

### http://localhost:3000/account/edit

image

### http://localhost:3000/signup

image

### http://localhost:3000/password/change

image

### http://localhost:3000/account/edit

image

## Types of changes

## Checklist:

### DIFF
--- a/templates/app/views/spree/user_passwords/edit.html.erb
+++ b/templates/app/views/spree/user_passwords/edit.html.erb
@@ -1,7 +1,7 @@
-<%= render partial: 'spree/shared/error_messages', locals: { target: @spree_user } %>
-
 <article class="auth-container" id="change-password">
   <h1><%= I18n.t('spree.change_my_password') %></h1>
+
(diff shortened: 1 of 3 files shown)


kennyadsl (approved):
Non-blocking but, what do you think about using `render partial:` more evenly in the project? Sometimes it was used with `partial` and `locals` and sometimes without them. Maybe it's a good time to choose an option and change it everywhere?

gsmendoza:
@kennyadsl Sounds good. I think I prefer the short-hand version of `render`. From https://api.rubyonrails.org/classes/ActionView/PartialRenderer.html#class-ActionView::PartialRenderer-label-Rendering+the+default+case:

image

I'll update the `render` calls everywhere on the site after this set of PRs, since it's likely to conflict with pending commits.

PR #200, opened 2021-11-04
Refactor user authentication pages

gsmendoza:
## Goal

As a solidus starter frontend user
I want the pages to be refactored 
So that they would be easier to extend and update

## How Has This Been Tested?

Manually tested. HTML diffs recorded in git logs.

### http://localhost:3000/login

image

### http://localhost:3000/checkout/registration

image

### http://localhost:3000/signup

image

### http://localhost:3000/password/recover

image

### http://localhost:3000/user/spree_user/password/edit?reset_password_token=xx4bQR48PfmzjSsaPkUP

image

### http://localhost:3000/account/edit

image

## Types of changes

- N/A: New feature (non-breaking change which adds functionality)

## Checklist:


templates/app/views/spree/user_passwords/edit.html.erb
```diff
@@ -1,7 +1,29 @@
 <article class="auth-container" id="change-password">
   <h1><%= I18n.t('spree.change_my_password') %></h1>
 
-  <%= render partial: 'spree/shared/error_messages', locals: { target: @spree_user } %>
+  <%= render 'spree/shared/error_messages', target: @spree_user %>
```
kennyadsl, line 4: Food for another PR but looks like a good candidate for a ViewComponent, WDYT?

kennyadsl (approved):
💘

templates/app/views/spree/user_passwords/edit.html.erb
gsmendoza: @kennyadsl Hmmm.. I would prefer to keep it as is because its logic is fairly simple. I think we would be making the partial more complicated if we wrapped it in a ViewComponent and/or we converted its html to `content_tag` blocks.

I would prefer prioritizing ViewComponent for complicated partials such as `templates/app/views/spree/components/search/_filter.html.erb`. 

Let me know if you have additional thoughts! :)

kennyadsl: The approach works for me, I'm sure you have more clear than me what we should prioritize!

PR #211, opened 2021-11-18
Fix account takeover through CSRF attack

gsmendoza:
(Copied from https://github.com/solidusio/solidus_auth_devise/commit/ec4fbe6876d1661a90976f5c03acf4805c36eb45)

This commit fixes an account takeover vulnerability when [Rails
`protect_from_forgery`](https://api.rubyonrails.org/classes/ActionController/RequestForgeryProtection/ClassMethods.html)
method is both:

- Executed whether as:
  - A `before_action` callback (the default)
  - A `prepend_before_action` (option `prepend: true`) before the
  `:load_object` hook in `Spree::UsersController` (most likely order to
  find).

- Configured to use `:null_session` or `:reset_session`
  strategies (`:null_session` is the default in case the no strategy is
  given, but `rails --new` generated skeleton use `:exception`).

Before this commit, the user was fetched in a `prepend_before_action`
hook named `:load_object`. I.e., the user was loaded into an instance
variable before touching the session as a safety countermeasure. As the
request went forward, `#update` was called on that instance variable.

The `:exception` strategy prevented the issue as, even if the user was
still fetched, the request was aborted before the update phase. On the
other hand, prepending `:protect_from_forgery` after the `:load_object`
hook (not very likely, as `ApplicationController` is loaded in
the first place and it's the most likely place to have that definition)
wiped the session before trying to fetch the user from it.

We could have fixed the most likely issue by just using a
`before_action` for `:load_object`, but it's safer not to rely on the
order of callbacks at all.

### DIFF
--- a/templates/app/controllers/spree/users_controller.rb
+++ b/templates/app/controllers/spree/users_controller.rb
@@ -2,12 +2,12 @@
 
 class Spree::UsersController < Spree::StoreController
   skip_before_action :set_current_order, only: :show, raise: false
-  prepend_before_action :load_object, only: [:show, :edit, :update]
   prepend_before_action :authorize_actions, only: :new
 
   include SolidusStarterFrontend::Taxonomies
 
   def show
+    load_object
     @orders = @user.orders.complete.order('completed_at desc')
   end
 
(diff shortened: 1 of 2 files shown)


waiting-for-dev (approved):
:+1:

PR #216, opened 2021-12-15
Apply demo styles to product page

gsmendoza:
## Goal

As a Solidus Starter Frontend maintainer
I would like the frontend to be consistent with the demo
So that people who have seen the demo won't be confused with the frontend provided by Solidus Starter Frontend

## Demo

https://drive.google.com/file/d/1xnZBefHYT7ty4wB6SD1M-4GZzdiqGa5Z/view?usp=sharing

### Solidus Demo

image

### Starter Frontend before changes

image

### Starter Frontend after changes

image

## Implementation strategy

The PR follows this strategy in copying the product page from Solidus Demo:

1. Refactor the product page code in preparation for the Solidus Demo replacement.
1. Copy the product page code from Solidus Demo with as little changes as possible. Override existing code.
2. Fix any issues that were caused by the Solidus Demo code. Restore overridden code if necessary.
3. Clean up code.

The PR splits the product page code into two main commits:

1. Copy product page layout from Solidus Demo.
2. Copy product form from Solidus Demo.

## Postponed code improvements 

In order to keep the PR as small as possible, I postponed improving some code in the PR. I have marked them with a comment in the PR. If you'd like them to be included in the PR, let me know :)

## Types of changes
- N/A: Bug fix (non-breaking change which fixes an issue)

## Checklist:


bin/sandbox
```diff
@@ -115,6 +115,8 @@ bundle exec rails generate solidus:auth:install --auto-run-migrations
 
 LOCATION='../template.rb' bundle exec rails app:template
 
+bundle binstub rspec-core
```
gsmendoza, line 118: Creates `bin/rspec` executable in sandbox directory.

templates/app/views/spree/products/_product_header.html.erb
```diff
@@ -0,0 +1,24 @@
+<header class="product-header">
+  <h1 class="product-header__title" itemprop="name">
+    <%= product.name %>
+  </h1>
+
+  <% if product.price_for_options(current_pricing_options)&.money and !product.price.nil? %>
```
gsmendoza, line 6: This price nil check is in Starter Frontend but not in the demo. The following spec fails if it is removed. However, the test failure only occurs in CI. I wasn't able to figure out why the failure doesn't appear locally.

templates/spec/system/products_spec.rb
```diff
@@ -166,7 +166,8 @@
 
     it 'displays price of first variant listed', js: true do
       click_link product.name
-      within("[data-js='price']") do
+
+      within("#product-price") do
```
gsmendoza, line 170: I wasn't able to figure out why the "Copy product form from Solidus Demo" commit caused this spec example to fail.

templates/app/assets/stylesheets/spree/frontend/base/_utilities.scss
```diff
@@ -0,0 +1,3 @@
+.d-hide {
+  display: none;
+}
```
gsmendoza, line 3: I placed the `utilities` SCSS file under base directory. It was under `app/assets/stylesheets/general` in Solidus Demo.

templates/app/assets/stylesheets/spree/frontend/components/products/_product_thumbnails.scss
```diff
@@ -5,3 +5,31 @@
   overflow: auto;
   padding: 0;
 }
+
+// Copied from Solidus Demo 3c11395ef6bfda0b906f68dbd11ab5ff567da747
```
gsmendoza, line 9: I postponed merging Solidus Demo CSS into the existing CSS for now.

templates/app/views/spree/products/_product_variants.html.erb
```diff
@@ -3,26 +3,32 @@
...
           <%= radio_button_tag(
             'variant_id',
             variant.id,
             index == 0,
-            data: { js: 'variant-radio', js_price: variant.price_for_options(current_pricing_options)&.money.to_s }
+            "data-js" => 'variant-radio',
+            "data-js_price" => variant.price_for_options(current_pricing_options)&.money.to_s,
+            "data-option-value-ids" => sorted_option_values(variant).to_json,
+            "data-target" => "variantOptions",
+            "data-price" => variant.price_for(current_pricing_options).to_html
```
gsmendoza, line 22: I postponed grouping the data properties.

templates/app/assets/javascripts/spree/frontend/product_selection_component.js
```diff
@@ -0,0 +1,103 @@
...
+}
+
+function updateView(variant) {
+  document.querySelector('#product-price').innerHTML = variant.dataset.price;
+}
+
+function selectVariant() {
+  this.variant = document.querySelector(`[data-option-value-ids="${JSON.stringify(this.currentSelection())}"]`);
+  if (this.variant) {
+    this.variant.click();
```
gsmendoza, line 88: I changed this from `this.variant.checked = true;` to `this.variant.click();` in order for the variant selection to refresh the product image. I'm not since if there's a better way of triggering this event.

templates/app/views/spree/products/product_selection/_option_type.html.erb
```diff
@@ -0,0 +1,30 @@
...
+            class: "selection-items",
+            type: "radio",
+            value: option_value.id,
+            checked: i.zero?,
+            name: "#{option_type.name}",
+            "data-action": "clickOnSelection",
+            "data-presentation": option_value.presentation,
+            "data-option-index": index
+          ) %>
+          <span class='<%= "selection-#{option_type.presentation.downcase}" %> <%= "selected-#{option_value.name.downcase}" %>'>
```
gsmendoza, line 23: I postponend converting this span HTML code into a Rails content tag.

templates/app/assets/stylesheets/spree/frontend/base/_utilities.scss
cpfergus1: I think this makes more sense, thank you for the correction

templates/app/helpers/product_options_helper.rb
```diff
@@ -0,0 +1,22 @@
...
+  def sorted_option_values(variant)
+    variant.option_values.sort_by { |value| value.option_type.position }.map(&:id)
+  end
+
+  def sorted_option_types(product)
+    product.option_types.sort_by(&:position)
+  end
+
+  # move to model
+  def option_values(option_type)
```
cpfergus1: I thought this method might make more sense to move into the product model, but never went back to move It. What are your thoughts on putting it into the product model?

templates/app/assets/javascripts/spree/frontend/product_selection_component.js
cpfergus1: Makes sense, great addition!

templates/app/helpers/spree/product_options_helper.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+module Spree
+  module ProductOptionsHelper
+    def product_variants_with_options
+      @product.variants_and_option_values_for(current_pricing_options)
```
kennyadsl, line 6: What about passing `product` as an argument to this method? It will be consistent with the other method above and I think it's better because we could use it in contexts where we don't have the instance variable (eg. product listing)

templates/app/helpers/spree/product_options_helper.rb
```diff
@@ -0,0 +1,24 @@
...
+      variant.option_values.sort_by { |value| value.option_type.position }.map(&:id)
+    end
+
+    def sorted_option_types(product)
+      product.option_types.sort_by(&:position)
+    end
+
+    # move to model
+    def option_values(option_type)
+      @product.variants.map do |variant|
```
kennyadsl, line 19: Same here for passing `product` as an argument.

templates/app/views/spree/products/_product_form.html.erb
```diff
@@ -15,7 +15,7 @@
     </div>
   <% end %>
   <%= render 'product_variants' %>
-  <% if @product.price and !@product.price.nil? %>
+  <% if @product.price %>
```
kennyadsl: ~On the next line we are using `product` instead of `@product`, what about using the same variable here?~

Saw you did it in the next commit 💌

templates/app/assets/stylesheets/spree/frontend/components/products/_product_page.scss
```diff
@@ -29,3 +29,50 @@
     }
   }
 }
+
+// Copied from Solidus Demo 3c11395ef6bfda0b906f68dbd11ab5ff567da747
```
kennyadsl, line 33: I'd remove these comments, not sure why it's needed (and we can move this information in the PR description, a git blame should be enough to retrieve them).

kennyadsl (commented):
Nice work, left some questions/comments. 👍

templates/app/helpers/product_options_helper.rb
gsmendoza: @cpfergus1 Hmmm.. Generally, it's good practice to limit access to a method only to the code that depend on it. As we broaden access to a method, we increase its cost of maintenance, since we have to check a wider area of code whenever we need to change the method.

In the case of the `option_values` method, if it can be reused in other areas of the codebase such as in the models or in the controllers, then yeah, it would make sense to move it to product model. 

However, at the moment, it seems like the method is only used in `product_selection/_option_type.html.erb`. Its logic is also very straightforward; we can get away without testing it :D Because of these factors, I think it might actually be better to inline the method in `product_selection/_option_type.html.erb`.

What do you think?

templates/app/helpers/spree/product_options_helper.rb
gsmendoza: Sounds good. I'll update the methods to receive `product` as an argument. TODO.

templates/app/assets/stylesheets/spree/frontend/components/products/_product_page.scss
gsmendoza: * I'd like to record where the code came from, lest people think I wrote it by myself lolz :) I think it might be best to record that in the commit message. TODO.
* I feel uncomfortable leaving the copied CSS unmerged with the existing CSS because without the comment, it won't be immediately obvious why they are separate. However, I realize that leaving the comment is a half-measure also, because in order to maintain the correctness of the comment, one would have to avoid editing the copied CSS :p So it looks like the best solution is to merge the existing and copied CSS in this PR. I'll look into it. TODO.

kennyadsl (approved):
Thanks @gsmendoza !!!

PR #229, opened 2022-06-03
Fix noise and deprecation warnings when running specs

gsmendoza:
Fixes three messages that appear when running the specs:

* Unexpected inner loop exception: unknown keyword: type.
* `price_for` is deprecated and will be removed.
* `rendered_component` is deprecated and will be removed in v3.0.0.

### DIFF
--- a/template.rb
+++ b/template.rb
@@ -57,7 +57,7 @@ end
 def add_spec_gems
   gem_group :development, :test do
     gem 'rspec-rails'
-    gem 'apparition', '~> 0.6.0'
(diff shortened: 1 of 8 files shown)


gsmendoza:
@waiting-for-dev I'll update this PR to use `page` instead. From https://github.com/github/view_component/pull/1380/files#diff-d66a35a69a27f26d54c6db1034c0a6b61e5df8df60e12c7030d97fe28faae9e2R37-R38:

```
        "`rendered_component` is deprecated and will be removed in v3.0.0. " \
        "Use `page` instead."
```

PR #231, opened 2022-07-18
Use `spree_current_user` instead of `try_spree_current_user`

gsmendoza:
## Description

This fixes SolidusStarterFrontend after the changes in
https://github.com/solidusio/solidus/pull/3923 (Deprecate
`try_spree_current_user`).

The change reveals some incorrect expections in the checkout specs. Please see the commit messages for details.

## Types of changes

- N/A - New feature (non-breaking change which adds functionality)
- N/A - Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- N/A - My change requires a change to the documentation.
- N/A - I have updated the documentation accordingly.

### DIFF
--- a/templates/spec/system/checkout_spec.rb
+++ b/templates/spec/system/checkout_spec.rb
@@ -103,8 +103,8 @@ RSpec.describe 'Checkout', type: :system, inaccessible: true do
 
       before do
         allow_any_instance_of(Spree::CheckoutController).to receive_messages(current_order: order)
-        allow_any_instance_of(Spree::CheckoutController).to receive_messages(try_spree_current_user: user)
-        allow_any_instance_of(Spree::OrdersController).to receive_messages(try_spree_current_user: user)
+        allow_any_instance_of(Spree::CheckoutController).to receive_messages(spree_current_user: user)
+        allow_any_instance_of(Spree::OrdersController).to receive_messages(spree_current_user: user)
 
         add_mug_to_cart
         click_button "Checkout"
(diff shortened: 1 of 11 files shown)


templates/spec/system/checkout_spec.rb
```diff
@@ -364,7 +364,7 @@
       click_on "Place Order"
 
       order = Spree::Order.last
-      expect(page).to have_current_path(spree.token_order_path(order, order.guest_token))
+      expect(page).to have_current_path(spree.order_path(order))
```
gsmendoza, line 367: From the commit message:

Expected behavior for the specs
----------------------------------

Given I am signed in
And I am checking out an order
When I finalize the order (click on "Place Order")
Then I should be redirected to the order path

Actual behavior of the broken specs
-----------------------------------

I'm redirected to the guest order path.

Cause
-----

`Spree::CheckoutController#completion_route` uses `spree_current_user`
to determine the order completion route, which is the order path if
there is a current user, and the guest order path if there is none.
However, the specs were previously stubbing only
`try_spree_current_user` and not `spree_current_user`. As a result, the
specs was redirecting incorrectly to the guest order path.

Solution
--------

The spec bugs were revealed in the previous commit, where all
`try_spree_current_user` references were changed to
`spree_current_user`. This commit fixes the expected order paths for the
specs.

waiting-for-dev (approved):
Thanks, @gsmendoza!

PR #237, opened 2022-08-15
Point solidus_compare to solidus_frontend repo

gsmendoza:
## Summary

* I've verified that SolidusStarterFrontend matched Solidus's frontend directory right before the directory was deleted. I did this by comparing SolidusStarterFrontend to Solidus at commit [935c8aeaaa2e427b6ea8f53d220b99aef2d98bf3](https://github.com/solidusio/solidus/commit/935c8aeaaa2e427b6ea8f53d220b99aef2d98bf3).
* https://github.com/solidusio/solidus_frontend/commit/2446d3bf420cfdf9aa94c80de4801ca0d476cdbb also ensures that SolidusFrontend matched Solidus's frontend directory. This means SolidusStarterFrontend matches SolidusFrontend as well.
* Pointing `solidus_compare` to the SolidusFrontend repo caused the diff hashes in the `solidus_compare`'s config to change. I think the hash changes were caused by the changes in the source directory values of `solidus_compare`'s config.

## Types of changes

## Checklist:

### DIFF
--- a/bin/solidus_compare
+++ b/bin/solidus_compare
@@ -101,7 +101,9 @@ end
 
 def solidus_compare
   report = @config['project_paths'].map do |path|
-    remote_source = "remotes/#{@config['source_name']}/#{@config['source_branch']}"
+    remote_source =
+      @config['source_commit'] || "remotes/#{@config['source_name']}/#{@config['source_branch']}"
+
     remote_path = "#{@config['source_base_path']}#{path}"
     result = `git diff "#{remote_source}" -- "#{remote_path}" "#{@config['local_base_path']}#{path}"`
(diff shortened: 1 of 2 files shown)


waiting-for-dev:
Thanks, @gsmendoza! I wonder if we still need to compare both repositories. If solidus_frontend is now deprecated, why do we need to keep comparing both? My understanding is that new changes should only be added to solidus_starter_frontend. Thoughts, @gsmendoza, @kennyadsl, @aldesantis?

gsmendoza:
@waiting-for-dev Hmmm.. You're right: from this point forward, new changes should only be added to `solidus_starter_frontend`, so there's no more need to ensure that it's in sync with `solidus_frontend`. I'll close this PR.

PR #236, opened 2022-08-08
Update installation instructions

waiting-for-dev:
## Description

We need to reflect that solidus_starter_frontend is now installed by
default by Solidus.

See solidusio/solidus#4490

## Motivation and Context
N/A

## How Has This Been Tested?
N/A

## Screenshots (if appropriate):

N/A

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:
- [ ] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.

PR #240, opened 2022-08-17
Remove configuration options from solidus_starter_frontend

gsmendoza:
## Dependencies

This is blocked by https://github.com/solidusio/solidus_starter_frontend/pull/236. Both PRs update the README. There is no actual conflict, but it may be better to see the README changes of this PR in the context of the README changes in #236.

## Description

`solidus_starter_frontend` currently exposes a [configuration file](https://github.com/solidusio/solidus_starter_frontend/blob/master/templates/lib/solidus_starter_frontend_configuration.rb) which controls some behavior in the generated templates.

Since the whole frontend is just copied into the main app, having a configuration class is perfectly useless and a bit confusing, because users can just edit the generated views manually.

## Types of changes

## Checklist:


templates/spec/controllers/controller_helpers_spec.rb
```diff
@@ -1,31 +0,0 @@
...
-  end
-
-  after do
-    I18n.reload!
-    I18n.locale = :en
-    I18n.enforce_available_locales = true
-  end
-
-  # Regression test for https://github.com/spree/spree/issues/1184
-  it "sets the default locale based off SolidusStarterFrontend::Config[:locale]" do
```
gsmendoza, line 26: Since `StoreController#config_locale` is directly set to `I18n.locale`, this spec is no relevant and can be deleted.

templates/app/views/spree/checkout/steps/_confirm_step.html.erb
```diff
@@ -7,22 +7,18 @@
...
-      name: :commit,
-      disabled: requires_acceptance
-    ) %>
-  <% end %>
+  <%= button_tag(
+    I18n.t('spree.place_order'),
+    class: 'button-primary',
+    name: :commit,
+    disabled: true
+  ) %>
```
gsmendoza, line 21: @aldesantis The terms and conditions check box was previously disabled by default. As part of removing the terms and conditions preference, I had two choices:

1. Follow the previous default behavior and delete the check box from the form, or
2. Change the form's behavior and show the check box.

I chose the second option because

1. It's easier for a Solidus developer to delete the check box than to re-add it.
2. I googled "checkout terms and conditions" and it seems that having a terms and conditions checkbox at the checkout is a common requirement. See for example, https://community.shopify.com/c/shopify-design/how-to-add-agreement-checkbox-of-terms-and-conditions-before/td-p/443197 and https://support.thrivecart.com/help/adding-terms-conditions-to-your-checkout-page/.

The downside of taking the second option is that I had to enable JavaScript on some checkout specs in order to enable the Place Order button once the terms and conditions check box is clicked.

Let me know your thoughts!

waiting-for-dev (approved):
LGTM!

templates/app/views/spree/checkout/steps/_confirm_step.html.erb
aldesantis: I think enabling it by default is reasonable 👍

aldesantis (approved):

PR #243, opened 2022-09-06
Remove Spree namespace from SolidusStarterFrontend Ruby code

gsmendoza:
## Types of changes

## Checklist:


template.rb
```diff
@@ -106,4 +114,8 @@ def print_security_update_message
   print_wrapped set_color(message.gsub("\n", ' '), :yellow)
 end
 
+def quick_update?
+  ENV.fetch('QUICK_UPDATE', nil)
+end
```
gsmendoza, line 119: I added a `QUICK_UPDATE` config to the template which, when used with the `bin/quick_update_script`, makes it easy to quickly update the sandbox during development.

templates/spec/solidus_starter_frontend_helper.rb
```diff
@@ -56,4 +56,6 @@
 
     ActionController::Base.cache_store = original_cache_store
   end
+
+  config.include SolidusStarterFrontend::TestingSupport::ControllerRequests, type: :controller
```
gsmendoza, line 60: This is needed for the controller specs to find the application routes. Otherwise, the controller specs would get the routes from the Spree engine through the following call sequence:

* `require 'solidus_dev_support/rspec/feature_helper'` (`spec/solidus_starter_frontend_helper.rb`)
* `require 'solidus_dev_support/rspec/rails_helper'` (`solidusio/solidus_dev_support/lib/solidus_dev_support/rspec/feature_helper.rb`)
* `config.include Spree::TestingSupport::ControllerRequests, type: :controller` (`solidusio/solidus_dev_support/lib/solidus_dev_support/rspec/rails_helper.rb`)

template.rb
```diff
@@ -64,11 +67,16 @@ def add_spec_gems
 end
 
 def copy_solidus_starter_frontend_files
+  remove_file 'app/views/layouts/application.html.erb'
```
gsmendoza, line 70: The template now forcefully removes any application layout from the Rails app in order to replace it with the application layout from Solidus.

templates/app/views/checkout/registration.html.erb
```diff
@@ -1,7 +1,7 @@
 <%= render 'spree/shared/error_messages', target: @user %>
```
gsmendoza, line 1: We can consider copying the `error_messages` partial from Spree. This will make it easier for a SolidusStarterFrontend user to modify it. Let me know what you think.

templates/app/helpers/spree/base_helper.rb
```diff
@@ -0,0 +1,154 @@
+# frozen_string_literal: true
+
+require 'carmen'
+
+module Spree
+  module BaseHelper
```
gsmendoza, line 6: I copied `Spree::BaseHelper` from Solidus Core in order to override some methods in the helper that call spree routes e.g. the `logo` helper which originally called `spree.root_path`. I wasn't able to figure out a way to override only those helpers used by the frontend. I tried the following:

1. Overriding `logo` in ApplicationHelper.
2. Overriding `logo` in a new local BaseHelper and setting `helper 'base'` in StoreController.

It appears the `Spree::Base` helper from the Solidus engine would always have higher precedence over local helpers from the app.

When copying `Spree::BaseHelper`, I had to override all the methods because once the `Spree::BaseHelper` module is overridden locally, it seems the app would no be longer to find non-overridden methods from the `Spree::BaseHelper` in the Solidus gem.

The only alternative I can think of is to replace the `Spree::Base` helpers in with local variations, something like:

```
module ApplicationHelper
    def app_logo(image_path = Spree::Config[:logo])
      link_to image_tag(image_path), root_path
    end
end
```

Let me know what you think!

templates/config/initializers/devise.rb
```diff
@@ -0,0 +1,3 @@
+Devise.setup do |config|
+  config.mailer = 'UserMailer'
+end
```
gsmendoza, line 3: I added a new initializer to override the Devise mailer and set it to the local UserMailer.

templates/config/routes.rb
```diff
@@ -1,15 +1,15 @@
...
     class_name: 'Spree::User',
     controllers: {
-      sessions: 'spree/user_sessions',
-      registrations: 'spree/user_registrations',
-      passwords: 'spree/user_passwords',
-      confirmations: 'spree/user_confirmations'
+      sessions: 'user_sessions',
+      registrations: 'user_registrations',
+      passwords: 'user_passwords',
+      confirmations: 'user_confirmations'
```
gsmendoza, line 12: Currently, this config produces Devise routes that still have `spree_user` in the URLs:

```
                                  Prefix Verb   URI Pattern                                                                                       Controller#Action
                                    root GET    /                                                                                                 home#index
                  new_spree_user_session GET    /user/spree_user/sign_in(.:format)                                                                user_sessions#new
                      spree_user_session POST   /user/spree_user/sign_in(.:format)                                                                user_sessions#create
              destroy_spree_user_session DELETE /user/spree_user/logout(.:format)                                                                 user_sessions#destroy
                 new_spree_user_password GET    /user/spree_user/password/new(.:format)                                                           user_passwords#new
                edit_spree_user_password GET    /user/spree_user/password/edit(.:format)                                                          user_passwords#edit
                     spree_user_password PATCH  /user/spree_user/password(.:format)                                                               user_passwords#update
                                         PUT    /user/spree_user/password(.:format)                                                               user_passwords#update
                                         POST   /user/spree_user/password(.:format)                                                               user_passwords#create
          cancel_spree_user_registration GET    /user/spree_user/cancel(.:format)                                                                 user_registrations#cancel
             new_spree_user_registration GET    /user/spree_user/sign_up(.:format)                                                                user_registrations#new
            edit_spree_user_registration GET    /user/spree_user/edit(.:format)                                                                   user_registrations#edit
                 spree_user_registration PATCH  /user/spree_user(.:format)                                                                        user_registrations#update
                                         PUT    /user/spree_user(.:format)                                                                        user_registrations#update
                                         DELETE /user/spree_user(.:format)                                                                        user_registrations#destroy
                                         POST   /user/spree_user(.:format)                                                                        user_registrations#create
```

I tried changing the config to

```
  devise_for(:user, {
    class_name: 'Spree::User',
    controllers: {
      sessions: 'user_sessions',
      registrations: 'user_registrations',
      passwords: 'user_passwords',
      confirmations: 'user_confirmations'
    },
    skip: [:unlocks, :omniauth_callbacks],
    path_names: { sign_out: 'logout' }
  })
```

However, this causes the specs to fail. It seems to break `spree_current_user`. I still need to investigate this.

Considering that the frontend ignores these Devise URLs in favor of custom URLs (`get '/login'`), I think it's tolerable at the moment to keep the `spree_user` in these Devise routes.

Let me know what you think!

templates/spec/controllers/checkout_controller_spec.rb
```diff
@@ -2,7 +2,7 @@
 
 require 'solidus_starter_frontend_helper'
```
gsmendoza, line 3: I kept `solidus_starter_frontend_helper` because it provides it adds our opinionated RSpec configuration to the app. It might be better to keep that configuration isolated from any RSpec configurations that the app's developer adds to the test suite. Let me know if you think that it makes more sense to merge this helper's configuration to the app's `rails_helper` config.

templates/config/routes.rb
kennyadsl: I think this may be solved with the `path` options: https://www.rubydoc.info/github/heartcombo/devise/main/ActionDispatch/Routing/Mapper%3Adevise_for. Let me know!

kennyadsl (commented):
Left some preliminary comments, thanks for the hard work here.

template.rb
gsmendoza: @kennyadsl Okay, I removed the commit for the quick update feature. Will submit a separate PR for it :+1:

templates/spec/solidus_starter_frontend_helper.rb
gsmendoza: I haven't tried namespacing each of the URL helper calls. However, I think it should be okay to override the routes in every spec file, like this:

```ruby
RSpec.describe CheckoutController, type: :controller do
  routes { Rails.application.routes }
end
```

Would you prefer this approach over overriding all of the controller specs only once in `solidus_starter_frontend_helper.rb`?

templates/config/routes.rb
gsmendoza: I'll give it one more try! TODO

templates/app/mailers/user_mailer.rb
```diff
@@ -1,17 +1,15 @@
 # frozen_string_literal: true
 
-module Spree
-  class UserMailer < BaseMailer
-    def reset_password_instructions(user, token, *_args)
-      @store = Spree::Store.default
-      @edit_password_reset_url = spree.edit_spree_user_password_url(reset_password_token: token, host: @store.url)
-      mail to: user.email, from: from_address(@store), subject: "#{@store.name} #{I18n.t(:subject, scope: [:devise, :mailer, :reset_password_instructions])}"
-    end
+class UserMailer < BaseMailer
```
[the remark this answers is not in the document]
gsmendoza: It's fixed in the later "Re-scope Base classes under Spree namespace" commit.

templates/app/controllers/taxons_controller.rb
```diff
@@ -1,30 +1,28 @@
 # frozen_string_literal: true
 
-module Spree
-  class TaxonsController < Spree::StoreController
-    helper 'spree/products', 'spree/taxon_filters'
+class TaxonsController < Spree::StoreController
```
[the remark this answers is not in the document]
gsmendoza: Yes, that's fixed in the "Remove Spree namespace from local controller names" commit.

templates/app/controllers/store_controller.rb
```diff
@@ -5,6 +5,8 @@ class StoreController < Spree::BaseController
   include Spree::Core::ControllerHelpers::Order
   include SolidusStarterFrontend::Taxonomies
 
+  layout 'application'
```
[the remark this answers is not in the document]
gsmendoza: I think I considered naming the layout differently, though I ultimately went with `application` since that's the default layout that comes with Rails. That being said, if we name the layout `storefront`, we can avoid deleting any application layout already present in the app. I think the less we force changes on the app, the more robust the template will be. What do you think?

templates/app/views/checkout/edit.html.erb
```diff
@@ -1,5 +1,5 @@
 <article class="checkout">
-  <%= render "shared/error_messages", target: @order %>
+  <%= render "spree/shared/error_messages", target: @order %>
```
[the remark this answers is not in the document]
gsmendoza: I think it's simple enough that I can rebase that change here. TODO

gsmendoza:
@kennyadsl Appreciate if you can look at this PR again. I made the following updates:

* Removed the Quick Update feature from the template. Will submit on a separate PR.
* Copied error messages partial from Solidus.
* Removed spree references from the Devise routes.

templates/app/helpers/spree/base_helper.rb
kennyadsl: > However, the biggest downside with this approach is that we can only remove these helpers on a major release, since they are considered public API. Is that correct?

Yes, but we can deprecate and stop using them before the removal.

gsmendoza: @kennyadsl Just to clarify: if we just deprecate the helpers in `Spree::BaseHelpers` in Solidus Core, we'd still need to keep the local BaseHelpers module under the Spree namespace in order to allow them to override the Solidus Core helpers. It is only when we actually delete the Solidus Core helpers that we can finally remove local BaseHelpers module from the Spree namespace.

templates/app/controllers/store_controller.rb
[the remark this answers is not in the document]
kennyadsl: > I think the less we force changes on the app, the more robust the template will be. What do you think?

I agree!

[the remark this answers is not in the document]
gsmendoza: Okay, I'll change the layout to `storefront`. TODO.

templates/app/controllers/store_controller.rb
```diff
@@ -3,7 +3,7 @@
 class StoreController < Spree::BaseController
   include Spree::Core::ControllerHelpers::Pricing
   include Spree::Core::ControllerHelpers::Order
-  include SolidusStarterFrontend::Taxonomies
+  include Taxonomies
```
elia, line 6: Do we actually need the `#taxonomies` helper in the devise controllers? can we just inline the method where it's used?

Maybe out of the scope, but give this is no more library code, but rather a reference implementation I think we can be more specific and less "future-proof", thoughts?

elia (approved):
❤️ can't wait to see this merged!

kennyadsl (approved):
👍 we can go ahead and merge this! Feel free to do it when you are ready @gsmendoza.

templates/spec/solidus_starter_frontend_helper.rb
gsmendoza: @elia Okay, let's go with inlining `solidus_dev_support/rspec/feature_helper` and removing the `config.include Spree::TestingSupport::ControllerRequests, type: :controller` line. TODO.

template.rb
gsmendoza: Yep, I'll remove this `remove_file` call once I rename the layout to `storefront`. TODO.

templates/app/controllers/store_controller.rb
gsmendoza: I'll check if Taxonomies is used in the Devise controllers. If not, we can inline the module in `Spree::BaseController`. TODO

gsmendoza: @elia I found that the Devise controllers call the `taxonomies` helper as follows:

* Devise controllers load the storefront layout.
* The storefront layout renders the `header` partial.
* The `header` calls `taxonomies`.

Viewed from a Devise controller page, the `taxonomies` helper is called in the headers section, highlighted in red below:

image

gsmendoza:
@kennyadsl @elia Appreciate if you can take one more look. I made the following updates:

- Inlined `solidus_dev_support/rspec/feature_helper`
- Removed `Spree::TestingSupport::ControllerRequests` from being included.
- Renamed application layout to storefront.

I also confirmed that `#taxonomies` helper is used in the Devise controllers, so I postponed making additional refactorings with that helper.

You can find the changes in the following commits:

* Inline solidus_dev_support/rspec/feature_helper.rb
* Inline solidus_dev_support/rspec/rails_helper.rb
* Do not require 'spree/testing_support/controller_requests
* Reorganize solidus_starter_frontend_helper.rb
* Rename spree_application layout to storefront
* Point StoreController to storefront layout
* Set layout to storefront in Devise controllers
* Expect default layout to be storefront
* No longer test changing the layout via the Solidus config

template.rb
```diff
@@ -89,6 +89,9 @@ def copy_solidus_starter_frontend_files
   copy_file 'config/initializers/solidus_auth_devise_unauthorized_redirect.rb'
   copy_file 'config/initializers/canonical_rails.rb'
 
+  copy_file 'config/initializers/devise.rb', 'tmp/devise.rb'
+  append_file 'config/initializers/devise.rb', File.read('tmp/devise.rb')
```
elia, line 93: nit, probably for another PR: I think we could have used https://github.com/rails/thor/blob/v1.2.1/lib/thor/actions/inject_into_file.rb#L26-L33 here as it supports injecting into existing blocks

template.rb
```diff
@@ -90,8 +90,15 @@ def copy_solidus_starter_frontend_files
...
-    Rails.autoloaders.main.ignore(Rails.root.join('app/monkey_patches'))
+    if Rails.autoloaders.main
+      # Rails 7
+      Rails.autoloaders.main.ignore(Rails.root.join('app/monkey_patches'))
+    else
+      # Rails 6 and older
+      config.after_initialize do
+        Rails.autoloaders.main.ignore(Rails.root.join('app/monkey_patches'))
+      end
+    end
```
elia, line 104: nit: Actually here we could even keep the if outside of `application.rb` since the template ran on an existing app with a definite Rails version:

```rb
if Rails.autoloaders.main # Or `Rails.gem_version >= Gem::Version.new('7')`
  application <<~RUBY
    # Load monkey patches
    Rails.autoloaders.main.ignore(Rails.root.join('app/monkey_patches'))
  RUBY
else
  application <<~RUBY
    # Load monkey patches
    config.after_initialize do
      Rails.autoloaders.main.ignore(Rails.root.join('app/monkey_patches'))
    end
  RUBY
end

File: templates/app/controllers/user_confirmations_controller.rb [hunk 16]
```diff
@@ -1,17 +1,19 @@
...
   helper 'spree/base', 'spree/store'
 
   include Spree::Core::ControllerHelpers::Auth
   include Spree::Core::ControllerHelpers::Common
   include Spree::Core::ControllerHelpers::Order
   include Spree::Core::ControllerHelpers::Store
-  include SolidusStarterFrontend::Taxonomies
+  include Taxonomies
+
+  layout 'storefront'
```
elia, line 13: Here and in the other devise controllers I think a less repetitive solution could be to set the `parent_controller` to `StoreController` in the devise configuration, unless there are reasons not to do so of course

elia (approved):
Left a few comments that can inspire further improvements, probably in followup PRs, but otherwise this was epic! 🙌

template.rb
gsmendoza: Yeah, this looks better than having that conditional written on the `application.rb` file :+1:

PR #234, opened 2022-07-23
Fix controllers

k1r8r0wn:
## Description

Fix several uninitialized constant errors.

## How Has This Been Tested?

Locally.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #246, opened 2022-09-08
Remove Spree namespace from SolidusStarterFrontend assets

gsmendoza:
## Background

Since the starter frontend is copy-pasted into the host application, we don't need to keep the Spree namespace in the generated files. Instead, we can just copy everything at the root level.

## Types of changes

## Checklist:


elia (approved):
Maybe for another PR, but we should decide how to namespace the project, be it `solidus-starter-fronted`, `storefront`, or `frontend`, in any case consistency would be great

kennyadsl (commented):
Why not getting rid of the `frontend/` part as well? Is it needed?

gsmendoza:
> Why not getting rid of the `frontend/` part as well? Is it needed?

No strong preference for keeping or removing `frontend/`. Let's see what it looks like without `spree/frontend`. Todo.

templates/app/assets/javascripts/solidus_starter_frontend.js
```diff
@@ -1,10 +1,11 @@
...
 // the installer will append this file to the app vendored assets here: vendor/assets/javascripts/spree/frontend/all.js'
 
-// This is generated by `solidus:install` at
-// `vendor/assets/javascripts/spree/frontend/all.js`. See `setup_assets` at
+// spree/frontend/all points to the
+// `vendor/assets/javascripts/spree/frontend/all.js` file generated by
+// `solidus:install`. See `setup_assets` at
 // https://github.com/solidusio/solidus/blob/master/core/lib/generators/solidus/install/install_generator.rb
-//= require spree/frontend/all
+// = require spree/frontend/all
```
elia, line 8: Not sure about this, but does the space stops it from being loaded or is just fo alignment?

gsmendoza: Thanks for the catch @elia ! I'll remove the space.

templates/app/assets/javascripts/solidus_starter_frontend.js
```diff
@@ -0,0 +1,16 @@
+// Placeholder manifest file.
+// the installer will append this file to the app vendored assets here: vendor/assets/javascripts/spree/frontend/all.js'
+
+// spree/frontend/all points to the
+// `vendor/assets/javascripts/spree/frontend/all.js` file generated by
+// `solidus:install`. See `setup_assets` at
+// https://github.com/solidusio/solidus/blob/master/core/lib/generators/solidus/install/install_generator.rb
+//= require spree/frontend/all
+
+//= require spree
```
kennyadsl, line 10: This seems the right place to get rid of this `spree` name as well. WDYT? Also, the description above could be probably changed to remove all the spree references.

templates/vendor/assets/javascripts/spree/frontend/all.js
```diff
@@ -5,5 +5,4 @@
 // the compiled file.
```
kennyadsl, line 5: Can't we move these vendor files into non-spree folders?

kennyadsl (commented):
👍

elia:
@gsmendoza I added a reference to #234 to the description as that PR should be unnecessary once this is merged 🙌

templates/app/assets/javascripts/solidus_starter_frontend.js
gsmendoza: @kennyadsl Yep, I have just checked `core/app/assets/javascripts/spree.js.erb` in Solidus and I found that the Spree JS helpers there are either replaced or not used in Solidus Starter Frontend. I'll remove this require call.

templates/vendor/assets/javascripts/spree/frontend/all.js
gsmendoza: @kennyadsl I don't think we can because extensions generated with SolidusDevSupport inject their asset links to these vendor paths. Please see the `add_javascripts` and `add_stylesheets` methods in https://github.com/solidusio/solidus_dev_support/blob/6270b0ba37cae69b7a6d4fcac928e0b4e55fa212/lib/solidus_dev_support/templates/extension/lib/generators/%25file_name%25/install/install_generator.rb.tt.

kennyadsl: If I got it correctly, that is going to work only with the old frontend. When using the new one, ideally no more JS files will be injected this way. @elia what do you think? Too early to remove the support for this?

gsmendoza: I removed `require spree` and didn't encounter any errors when I tested checkout.

elia: TBH, I'm fine with both, on one side very few JS files from extensions will keep working on a front end HTML that deviates conspicuously from the original solidus_frontend, on the other we could try to support the installation process and then let people to fix their app by themselves.

elia (approved):
Re-approving, we'll probably remove all legacy JS hooks, but let's do it in small steps.

PR #247, opened 2022-09-08
Add solidus_auth_devise if missing in the target app

elia:
With this change the app template can be run on apps that were installed with either `--frontend=none` and `--frontend=solidus_frontend`

## Description

This adds support for:
- ~~removing `solidus_frontend` from the gemfile if it's there~~
- ~~replacing `solidus` with its subcomponents if it's there~~
- adding `solidus_auth_devise` to the gemfile if it's not there

This should be compatible with running it both from the `solidus:install` generator and standalone after the fact.
In order to support both migrations for solidus_auth_devise are not run.

## Motivation and Context

This started with an investigation in https://github.com/solidusio/solidus/discussions/4583.

## How Has This Been Tested?

```
rails new my_app --skip-javascript
cd my_app
bundle add solidus
bin/rails g solidus:install --frontend=none --with-authentication=false
git add .
git commit -m "vanilla rails with solidus"
# and then repeatedly
g reset --hard
g clean -fdx
bin/rails app:template LOCATION=~/Code/Nebulab/solidus_starter_frontend/template.rb --trace
bin/rails db:migrate
bundle exec rspec 
```

## Types of changes

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ We aim to deliver:
 - a reusable component based architecture
 - simple SASS styling strictly based on BEM
 - the elimination of jQuery as a dependency by rewriting frontend functionality
-in vanilla JavaScript
+  in vanilla JavaScript
 
 All of this while keeping and improving on the functionality of the current
 [Solidus][solidus] frontend subcomponent.
@@ -33,16 +33,14 @@ Just run:
 rails new store --skip-javascript
 cd store
 bundle add solidus
-bin/rails generate solidus:install --auto-accept
+bin/rails generate solidus:install --frontend=soldius_starter_frontend
 ```
 
 That will create a new Solidus application with SolidusStarterFrontend as its
 storefront.
 
-Please note that `--auto-accept` will add
-[Solidus Auth Devise](https://github.com/solidusio/solidus_auth_devise)
-to your application. SolidusStarterFrontend requires the application to include
-the gem.
+Please note that [Solidus Auth Devise](https://github.com/solidusio/solidus_auth_devise)
+will also be added to your application as it's required by SolidusStarterFrontend.
 
 ### For existing stores
 
@@ -68,10 +66,11 @@ You'll also need to make sure that
 [Solidus Auth Devise](https://github.com/solidusio/solidus_auth_devise)
 is installed in your application.
 
-You can then copy the starter frontend files to your project:
+Then you can run the app template with this command:
 
 ```shell
-$ LOCATION="https://raw.githubusercontent.com/solidusio/solidus_starter_frontend/main/template.rb" bin/rails app:template
+$ bin/rails app:template LOCATION="https://github.com/solidusio/solidus_starter_frontend/raw/master/template.rb"
+$ bin/rails db:migrate
 ```
 
 ## Considerations
--- a/template.rb
+++ b/template.rb
@@ -18,18 +18,18 @@ def add_template_repository_to_source_path
 
     tempdir = Dir.mktmpdir("solidus_starter_frontend-")
     repo_dir = tempdir
-
+    url_path = URI.parse(__FILE__).path
+    branch = url_path[%r{solidus_starter_frontend/(raw/)?(.+?)/template.rb}, 2]
+    owner = url_path[%r{/([^/]+)/solidus_starter_frontend/}, 1]
     at_exit { FileUtils.remove_entry(tempdir) }
 
     git clone: [
       "--quiet",
-      "https://github.com/solidusio/solidus_starter_frontend.git",
+      "https://github.com/#{owner}/solidus_starter_frontend.git",
       tempdir
     ].map(&:shellescape).join(" ")
 
-    if (branch = __FILE__[%r{solidus_starter_frontend/(.+)/template.rb}, 1])
-      Dir.chdir(tempdir) { git checkout: branch }
-    end
+    Dir.chdir(tempdir) { git checkout: branch } if branch
   else
     repo_dir = File.dirname(__FILE__)
   end
--- a/template.rb
+++ b/template.rb
@@ -40,10 +40,18 @@ def add_template_repository_to_source_path
 end
 
 def install_gems
+  add_solidus_auth_devise_if_missing
   add_solidus_starter_frontend_dependencies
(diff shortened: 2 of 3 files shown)


waiting-for-dev:
Hey @elia 👋 

I've tried running the template after installing Solidus with solidus_frontend from GitHub:

```
# Gemfile
gem "solidus", github: "solidusio/solidus"

# Bash
bin/rails g solidus:install --frontend=solidus_frontend
bin/rails app:template LOCATION="https://github.com/elia/solidus_starter_frontend/raw/elia/legacy_frontend_migration_support/template.rb"
```

And the command ends up failing with:

```
run  bundle add solidus_core solidus_backend solidus_api solidus_sample --version \>\=\ 0 --source https://github.com/solidusio/solidus.git --git https://github.com/solidusio/solidus.git --ref ''
Fetching https://github.com/solidusio/solidus.git
fatal: Needed a single revision
Git error: command `git rev-parse --verify ''` in directory /usr/local/bundle/cache/bundler/git/solidus-169f1ecb1aee2122950e6d586daf2410f62df66e has failed.
Revision  does not exist in the repository https://github.com/solidusio/solidus.git. Maybe you misspelled it?
If this error persists you could try removing the cache directory '/usr/local/bundle/cache/bundler/git/solidus-169f1ecb1aee2122950e6d586daf2410f62df66e'
```

Anyway, I'm not sure we should try to manage all that complexity from here. We're already doing it from `solidus_core`, and it's meant to be temporary until `solidus_frontend` is removed from the meta-gem. I think we'll be doing a favor to ourselves if we only maintain that from one place for the time being. I found out that Bundle is very confusing when it comes to modifying the Gemfile (see https://github.com/solidusio/solidus/pull/4504), so I'd leave the manual instructions here.

However, I'd extract some very useful work from here, like taking care of the installation of solidus_auth_devise when it's not present (see https://github.com/solidusio/solidus/blob/b11b8f77b78e7bbb0d0b684d5ebaa3e0cd204d98/core/lib/generators/solidus/install/install_generator/install_frontend.rb#L39) or the leftover reference to Nebulab.

What do you think about it?

elia:
@waiting-for-dev extracted the gem/bundler hackery from this PR and just left the things you pointed out, ready for another (review) 🎠 ride!

waiting-for-dev:
Hey @elia, thanks for working on it! A couple of questions/comments:

- Could you move your branch to the `solidusio/solidus_starter_frontend` repository? I've tried modifying the Solidus main installer to point to the template on your fork, but the installer complains with `error: pathspec 'elia/legacy_frontend_migration_support' did not match any file(s) known to git`. I haven't looked into it in detail, but I suspect it's related to: https://github.com/solidusio/solidus_starter_frontend/blob/a81ba0eca1c5b1843702476670c2ad11eeff67aa/template.rb#L15

- Why don't we run migrations automatically? I'm not against it; it's probably the most sensible default. But it's not consistent with [Solidus default](https://github.com/nebulab/solidus/blob/17f56b53b5e77e8e4212381ce41e08c605664067/core/lib/generators/solidus/install/install_generator.rb#L22). What are your thoughts on that?

elia:
Hey @waiting-for-dev, thanks for the review!

> Could you move your branch to the solidusio/solidus_starter_frontend repository?

I think I didn't have permission to create branches there, at least at the time, anyway if you want to modify it it should be possible from the command line with `gh pr checkout 247`

> I've tried modifying the Solidus main installer to point to the template on your fork

I'll try that and dig into the issue

> Why don't we run migrations automatically? I'm not against it; it's probably the most sensible default. But it's not consistent with [Solidus default](https://github.com/nebulab/solidus/blob/17f56b53b5e77e8e4212381ce41e08c605664067/core/lib/generators/solidus/install/install_generator.rb#L22). What are your thoughts on that?

Ironically I did it to stay consistent with the default because I noticed that they were held back in the install_generator, so I thought there might have been a reason for that
https://github.com/solidusio/solidus/blob/8052691e31282b986aca69b60e2ba82499c8e4b9/core/lib/generators/solidus/install/install_generator.rb#L162 – anyway no strings attached to that I'm fine with any direction we wanna take.

waiting-for-dev:
> I think I didn't have permission to create branches there, at least at the time, anyway if you want to modify it it should be possible from the command line with gh pr checkout 247

It looks like you're admin now 🙂 Anyway, I didn't know the trick about `gh pr checkout`, so it turns out I already have created it 😆 

> I'll try that and dig into the issue

I did, and it worked, but I think it's mixing content from `main` and the PR branch as it did run the solidus_auth_devise migrations 🤷 I haven't looked into it in more detail, but probably it's not possible to test it in a 100% realistic scenario until it's merged. Maybe we can try with a fork, though, updating the script accordingly.

>Ironically I did it to stay consistent with the default because I noticed that they were held back in the install_generator, so I thought there might have been a reason for that
https://github.com/solidusio/solidus/blob/8052691e31282b986aca69b60e2ba82499c8e4b9/core/lib/generators/solidus/install/install_generator.rb#L162 – anyway no strings attached to that I'm fine with any direction we wanna take.

I'm fine with either, too. Maybe it makes more sense to run the migrations just to be consistent with the main repo.

elia:
```
         run  git clone --quiet https://github.com/solidusio/solidus_starter_frontend.git /var/folders/hn/z7b7s1kj3js4k7_qk3lj27kr0000gn/T/solidus_starter_frontend-20220914-37484-1maaisw from "."
         run  git checkout elia/legacy_frontend_migration_support from "."
error: pathspec 'elia/legacy_frontend_migration_support' did not match any file(s) known to git
```

@waiting-for-dev the issue is the one you pointed out, setting the template to `https://raw.githubusercontent.com/elia/solidus_starter_frontend/elia/legacy_frontend_migration_support/template.rb` or   `https://github.com/elia/solidus_starter_frontend/raw/elia/legacy_frontend_migration_support/template.rb` should work fine now

elia:
@waiting-for-dev also migrations are now ran

waiting-for-dev:
Oh, it finally went through. A bit slow, but it doesn't look this PR changed anything about that.

waiting-for-dev (approved):

PR #250, opened 2022-09-13
Extract OrdersController#populate into RESTful action

gsmendoza:
## Background

This PR is part of a broader task to make actions from OrdersController RESTful.

## Previous implementations

See https://github.com/solidusio/solidus/pull/4102.

## Types of changes

## Checklist:


templates/spec/requests/spree/orders_spec.rb
```diff
@@ -6,81 +6,6 @@
   let!(:store) { create(:store) }
   let(:variant) { create(:variant) }
 
-  context "#populate" do
```
gsmendoza, line 9: I deleted the `populate` specs because 1) they display the deprecation warning and 2) they're already tested by the `OrderContentsController#create` specs.

templates/app/controllers/spree/orders_controller.rb
```diff
@@ -77,6 +77,8 @@ def populate
         end
       end
     end
+    deprecate populate: 'please use OrdersContentController#create instead',
```
kennyadsl, line 80: I think we can just remove this. Being a template, deprecating code is not needed.

kennyadsl (changes requested):

templates/app/controllers/spree/orders_controller.rb
gsmendoza: Hehe.. You're right. I'll remove the deprecation notice. TODO.

gsmendoza:
@kennyadsl I updated the PR

* Removed `populate_redirect`.
* Removed the deprecation commit.
* Squashed the commits replacing `OrdersController#populate` with `OrderContentsController#create`.

templates/app/controllers/spree/order_contents_controller.rb
```diff
@@ -0,0 +1,47 @@
+module Spree
```
waiting-for-dev, line 1: Missing `frozen-string-literals` comment 🙂

templates/config/routes.rb
```diff
@@ -46,11 +48,9 @@
   get '/checkout/:state', to: 'checkout#edit', as: :checkout_state
   get '/checkout', to: 'checkout#edit', as: :checkout
 
-  get '/orders/populate', to: 'orders#populate_redirect'
```
waiting-for-dev, line 49: No replacement for that?

waiting-for-dev (commented):
Thanks, @gsmendoza, some comments 🙂

templates/app/controllers/spree/order_contents_controller.rb
gsmendoza: Oops! Missed that one. Fixed!

gsmendoza:
@waiting-for-dev Updated the PR:

* Added the missing `frozen_string_literal` comment.

templates/config/routes.rb
kennyadsl: Yes, Solidus is not using this route. It was added because people may refresh the browser page when adding to the cart in some weird way (maybe this is not more possible with recent browsers?) causing a 404 to happen. I think we can avoid this.

kennyadsl (approved):
👍 I suspect some rebasing pain merging this before #246. Up to how how to handle the merge order!

waiting-for-dev (approved):
Thanks @gsmendoza !

PR #251, opened 2022-09-14
Embed solidus_dev_support spec helpers and remove the dependency

elia:
Albeit `solidus_dev_support` simplified a few things with the rspec setup it was never intended to be used in stores, but rather only to aid extension development.

## Description

- Embed what we needed for the current project from dev-support
- Remove support for older Solidus version

## Motivation and Context

This change was suggested by an issue raised in the `#support` channel on slack about the changelog generator dependency relying on a version of the `async` gem which is only compatible with Ruby 3.

## How Has This Been Tested?

Relying on the current suite.

## Types of changes

## Checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -31,6 +31,7 @@ commands:
         type: string
         default: |
           cd sandbox
+          bundle add rspec_junit_formatter --group test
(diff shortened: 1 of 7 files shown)


kennyadsl (approved):
Yeeeees

PR #252, opened 2022-09-15
Faster clones for the templates repo

elia:
Improve reliability and performance with a shallow clone.

## Description

Since the clone is quite disposable we can be efficient and do a shallow clone of the branch we're interested in.

We're assuming that when used locally we have a git repo available and that all files are committed.

## Motivation and Context

Having the code work both from a file and from a remote git clone can hide some bugs (since the CI will only run with local code). 

## How Has This Been Tested?

The CI now covers most of the git related code, since it's shared.

## Types of changes

## Checklist:

### DIFF
--- a/template.rb
+++ b/template.rb
@@ -13,27 +13,29 @@ end
 # invoked remotely via HTTP, that means the files are not present locally.
 # In that case, use `git clone` to download them to a local temporary dir.
 def add_template_repository_to_source_path
+  require "shellwords"
+  require "securerandom"
+
   if __FILE__ =~ %r{\Ahttps?://}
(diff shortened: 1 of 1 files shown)


kennyadsl (approved):
Nice! A bit scared by

> We're assuming that when used locally we have a git repo available and that all files are committed.

though. Do you think it could be a problem in some context?

elia:
@kennyadsl that use case is there mainly to aid development and CI testing, regular use would always go through a URL

waiting-for-dev:
That's a great improvement, thanks, @elia!

PR #253, opened 2022-09-16
Refactor SolidusStarterFrontend CircleCI jobs

gsmendoza:
Goal
----

As a SolidusStarterFrontend maintainer, I want to be able to 

* Clearly see the differences of SolidusStarterFrontend CircleCI jobs against each other.
* Test SolidusStarterFrontend against Ruby, Rails, and DB versions supported by the latest Solidus version.

Background
----------

<https://github.com/solidusio/solidus_starter_frontend/commit/4bebb8d0df761ddb86306bf585cf5a82d4def572> updated the Circle CI config to mix and match Ruby, Rails, and Solidus versions. However, with the introduction of different Ruby versions to the jobs, the job names no longer adequately describe the scenarios they run.

Proposed solution
-----------------

-   Define a `run-specs` job with the following default parameters:
    -   Solidus Branch: v3.2
    -   Rails version: 7.0
    -   Ruby version: 3.1
    -   Database: Postgres

-   Use [CircleCI job parameters](https://circleci.com/docs/jobs-steps#passing-parameters-to-jobshttps://circleci.com/docs/jobs-steps#passing-parameters-to-jobs) to call `run-specs` but with different parameters.

-   Create additional jobs for the following scenarios:
    -   Solidus branch
        -   master
    -   Rails
        -   ~> 6.1
    -   Ruby
        - 3.0.4
        - 2.7.6
    -   Database
        -   MySQL

## What about Rails 5.2?

I tried testing SSF with Rails 5.2, but I encountered the following issues. Please see the table below. In each row, I indicate the possible solution I tried to fix the issue:

| Solidus | Rails | Ruby  | Psych | Discard | Issue                                                                                                                                                             | Possible solution                                                        |
|---------|-------|-------|-------|---------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------|--------------------------------------------------------------------------|
| 3.2     | 5.2   | 3.0   | N/A   | N/A     | Rails 5.2 doesn't work with Ruby 3.0 · Issue [#40938](https://github.com/rails/rails/issues/40938) – GitHub                                                                                                      | Lower Ruby to 2.7.6                                                      |
| 3.2     | 5.2   | 2.7.6 | N/A   | N/A     | Getting “Psych::BadAlias: Cannot load database configuration: Unknown alias: default” when running bin/sandbox.                                                   | Lower Psych to < 4.0                                                     |
| 3.2     | 5.2   | 2.7.6 | 3.3.3 | N/A     | NameError: undefined local variable or method `acts_as_paranoid' for #<Class:0x000055b1b970d058> - Paranoia gem has been replaced by Discard gem in [Solidus 3.0.0](https://github.com/solidusio/solidus/blob/bdebdbc15457487ab412702a02980bbbe32d3a6e/CHANGELOG.md#solidus-300-v30-2021-04-20) | Lower Solidus to 2.11.                                                   |
| 2.11.17 | 5.2   | 2.7.6 | 3.3.3 | N/A     | NameError: uninitialized constant Spree::User::Discard                                                                                                            | Add discard gem                                                          |
| 2.11.17 | 5.2   | 2.7.6 | 3.3.3 | 1.2.1   | undefined method `autoloaders' for Rails:Module                                                                                                                   | **Haven't tried**: Monkey-patch Spree::BaseHelpers with classic Rails loading for Rails 5.2 |

Please see https://github.com/solidusio/solidus_starter_frontend/tree/gsmendoza/sol-429-refactor-solidusstarterfrontend-circleci--rails-5-2 for the working branch for Rails 5.2 support.

Because it's getting complicated to get SolidusStarterFrontend to work with Rails 5.2, and because Rails 5.2 has already reached [end of life](https://endoflife.date/rails), I intend to no longer add a Rails 5.2 job to the CircleCI config.

## Types of changes

## Checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -100,32 +100,27 @@ workflows:
   "Run specs on supported Solidus versions":
     jobs:
       - run-specs:
-          name: run-specs-with-postgres
-          database: 'postgres'
+          name: run-specs-with-solidus-3-2-rails-7-ruby-3-1-postgres
+
+      - run-specs:
+          name: run-specs-but-with-solidus-latest
+          solidus_branch: 'master'
+
+      - run-specs:
+          name: run-specs-but-with-rails-6
+          rails_version: '~> 6.1'
+
+      - run-specs:
+          name: run-specs-but-with-ruby-3-0
+          ruby_version: '3.0'
+
+      - run-specs:
+          name: run-specs-but-with-ruby-2
           ruby_version: '2.7'
-          solidus_branch: 'v3.1'
-          rails_version: '~> 6.1'
 
       - run-specs:
-          name: run-specs-with-mysql
+          name: run-specs-but-with-mysql
           database: 'mysql'
-          ruby_version: '3.0'
-          solidus_branch: 'v3.1'
-          rails_version: '~> 6.1'
-
-      - run-specs:
-          name: run-specs-with-postgres-and-solidus-latest
-          database: 'postgres'
-          ruby_version: '3.0'
-          solidus_branch: 'master'
-          rails_version: '~> 7.0'
-
-      - run-specs:
-          name: run-specs-with-mysql-and-solidus-latest
-          database: 'mysql'
-          ruby_version: '3.1'
-          solidus_branch: 'master'
-          rails_version: '~> 7.0'
 
   "Weekly run specs against master":
     triggers:
@@ -137,29 +132,24 @@ workflows:
                 - master
     jobs:
       - run-specs:
-          name: run-specs-with-postgres
-          database: 'postgres'
+          name: run-specs-with-solidus-3-2-rails-7-ruby-3-1-postgres
+
+      - run-specs:
+          name: run-specs-but-with-solidus-latest
+          solidus_branch: 'master'
+
+      - run-specs:
+          name: run-specs-but-with-rails-6
+          rails_version: '~> 6.1'
+
+      - run-specs:
+          name: run-specs-but-with-ruby-3-0
+          ruby_version: '3.0'
+
+      - run-specs:
+          name: run-specs-but-with-ruby-2
           ruby_version: '2.7'
-          solidus_branch: 'v3.1'
-          rails_version: '~> 6.1'
 
       - run-specs:
-          name: run-specs-with-mysql
+          name: run-specs-but-with-mysql
           database: 'mysql'
-          ruby_version: '3.0'
-          solidus_branch: 'v3.1'
-          rails_version: '~> 6.1'
-
-      - run-specs:
-          name: run-specs-with-postgres-and-solidus-latest
-          database: 'postgres'
-          ruby_version: '3.0'
-          solidus_branch: 'master'
-          rails_version: '~> 7.0'
-
-      - run-specs:
-          name: run-specs-with-mysql-and-solidus-latest
-          database: 'mysql'
-          ruby_version: '3.1'
-          solidus_branch: 'master'
-          rails_version: '~> 7.0'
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -72,55 +72,55 @@ commands:
...
+        default: 'postgres'
 
 workflows:
   "Run specs on supported Solidus versions":
     jobs:
-      - run-specs-with-postgres
-      - run-specs-with-mysql
-      - run-specs-with-postgres-and-solidus-latest
-      - run-specs-with-mysql-and-solidus-latest
+      - run-specs
```
elia, line 102: Maybe we could add a name to also describe this job, so that in the PRs status is immediately clear what's failing?

.circleci/config.yml
```diff
@@ -72,55 +72,55 @@ commands:
...
+          name: run-specs-but-with-ruby-3-0
+          ruby_version: '3.0'
+
+      - run-specs:
+          name: run-specs-but-with-ruby-2
+          ruby_version: '2.7'
+
+      - run-specs:
+          name: run-specs-but-with-mysql
+          database: 'mysql'
```
elia, line 122: nit, non-mandatory: have you considered using a matrix setup with exclusions, it also supports dynamically composing the names based on parameters https://circleci.com/docs/configuration-reference#matrix-requires-version-21

elia (approved):
Great to see more automatic coverage!

Left just a couple of nits, feel free to ignore them, update the PR and ask for another review, or handle them separately 🙌

.circleci/config.yml
```diff
@@ -130,7 +136,30 @@ workflows:
...
+          solidus_branch: 'v3.1'
+          rails_version: '~> 6.1'
+
+      - run-specs:
+          name: run-specs-with-mysql
+          database: 'mysql'
+          ruby_version: '3.0'
+          solidus_branch: 'v3.1'
+          rails_version: '~> 6.1'
+
```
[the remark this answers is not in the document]
kennyadsl: I think they are going to be removed in the next commit.

kennyadsl (approved):
👍 for not supporting Rails 5.2. I think it's not worth the effort considering that this will most likely be installed on the latest Rails version. 

I like the @elia suggestion of using the matrix, but I'm also ok with the current setup. Up to you!

Thanks @gsmendoza!

.circleci/config.yml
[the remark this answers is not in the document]
gsmendoza: Yes, this is updated in the [Test SolidusStarterFrontend against various Ruby, Rails, and DB versions](https://github.com/solidusio/solidus_starter_frontend/pull/253/commits/51f3ee9edf40d0c52b12aa79960e04cc880665c0) commit.

.circleci/config.yml
gsmendoza: Hmm... I was initially hesitant that naming the base job as `run-specs-on-solidus-3-2-rails-7-ruby-3-1-postgres` would be unwieldy, but it turns out that's it's kinda okay. I'll rename it to that.

waiting-for-dev (approved):
Thanks! Also in favor of Elia's suggestion, but it looks great regardless.

.circleci/config.yml
gsmendoza: @elia Cool, it's the first time for me to see CircleCI matrices. However, having looked at the docs, I'm hesitant to switch to using matrices for this build:

* If we create a matrix for the 2 Solidus branches, 2 Rails versions, 3 Ruby versions, and 2 databases, we'll end up with 24 jobs, which seems excessive. Furthermore, we still have some flaky specs in our test suite right now, so a build with that many jobs would be unstable.
* If we use matrices but exclude everything to arrive at the 6-job config that we have now, we'll be excluding 18 combinations. 

I suppose we can review this if we find ourselves adding jobs to the build with more  complicated combinations of the parameters.

Your thoughts?

PR #255, opened 2022-09-22
Refactor monkey-patches config

gsmendoza:
SolidusStarterFrontend has two kinds of monkey-patch configuration: one for Rails 7 and another for Rails 6. Currently, the kind of configuration ran is determined during boot time, specifically when the Solidus app loads `application.rb`. We can simplify apps generated with SolidusStarterFrontend by determining the configuration at frontend installation time, i.e. when we're installing SolidusStarterFrontend on the app.

Requirements
------------

**For a Rails 7 app, load monkey-patches immediately when application.rb is loaded**

Given my app is Rails 7

When I install Solidus with SolidusStarterFrontend on the app

Then it should add the following monkey-patch configuration to the `application.rb` file of the app:

```
class Application < Rails::Application
  # ...
  Rails.autoloaders.main.ignore(Rails.root.join('app/monkey_patches'))
  # ...
end
```

**For a Rails 6 app, load monkey-patches only after the app has initialized**

Given my app is Rails 6

When I install Solidus with SolidusStarterFrontend on the app

Then it should add the following monkey-patch configuration to the `application.rb` file of the app:

```
class Application < Rails::Application
  # ...
  config.after_initialize do
    Rails.autoloaders.main.ignore(Rails.root.join('app/monkey_patches'))
  end
  # ...
end
```

Background
----------

Suggested by @elia in
<https://github.com/solidusio/solidus_starter_frontend/pull/243#discussion_r974264368>.

## Types of changes

~- [ ] Bug fix (non-breaking change which fixes an issue)~
~- [ ] New feature (non-breaking change which adds functionality)~
~- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)~

## Checklist:

~- [ ] My change requires a change to the documentation.~
~- [ ] I have updated the documentation accordingly.~

### DIFF
--- a/template.rb
+++ b/template.rb
@@ -92,23 +92,33 @@ def copy_solidus_starter_frontend_files
   copy_file 'config/initializers/devise.rb', 'tmp/devise.rb'
   append_file 'config/initializers/devise.rb', File.read('tmp/devise.rb')
 
-  application <<~RUBY
-    if Rails.autoloaders.main
-      # Rails 7
+  if Rails.gem_version >= Gem::Version.new('7')
+    application <<~RUBY
       Rails.autoloaders.main.ignore(Rails.root.join('app/monkey_patches'))
-    else
-      # Rails 6 and older
+
+      config.to_prepare do
+        Dir.glob(Rails.root.join('app/monkey_patches/**/*_monkey_patch.rb')).each do |monkey_patch|
+          load monkey_patch
+        end
+      end
+
+    RUBY
+  else
+    application <<~RUBY
       config.after_initialize do
         Rails.autoloaders.main.ignore(Rails.root.join('app/monkey_patches'))
       end
-    end
 
-    config.to_prepare do
-      Dir.glob(Rails.root.join('app/monkey_patches/**/*_monkey_patch.rb')).each do |monkey_patch|
-        load monkey_patch
+      config.to_prepare do
+        Dir.glob(Rails.root.join('app/monkey_patches/**/*_monkey_patch.rb')).each do |monkey_patch|
+          load monkey_patch
+        end
       end
-    end
 
+    RUBY
+  end
+
+  application <<~RUBY
     if defined?(FactoryBotRails)
       initializer after: "factory_bot.set_factory_paths" do
         require 'spree/testing_support'


template.rb
```diff
@@ -92,23 +92,33 @@ def copy_solidus_starter_frontend_files
...
+    application <<~RUBY
       Rails.autoloaders.main.ignore(Rails.root.join('app/monkey_patches'))
-    else
-      # Rails 6 and older
+
+      config.to_prepare do
+        Dir.glob(Rails.root.join('app/monkey_patches/**/*_monkey_patch.rb')).each do |monkey_patch|
+          load monkey_patch
+        end
+      end
```
gsmendoza, line 103: I duplicated the `to_prepare` block in each branch of the conditional to ensure that it gets written together with the `Rails.autoloaders.main.ignore` block.

waiting-for-dev:
Thanks, @gsmendoza. I think we need to hold back a bit. I don't like we're adding code to autoload monkey patches when we deprecated that default behavior in solidus_core (see https://github.com/solidusio/solidus/pull/4231).

The overridden helpers don't seem to be used elsewhere in solidus_{core, backend, api}, so why not including them from somewhere else that is not `Spree::Base` and remove the monkey patches altogether?

elia:
If we want a more modern approach (i.e. zeitwerk based) we can grab the implementation from https://github.com/solidusio/solidus_support/pull/60 and use `.on_load("Spree::BaseHelper")` so that the code is future proof.

On the other side long term I think we should free Core from frontend helpers that are actually very specific to the legacy solidus_frontend, and push them all down in SSF and SF (BaseHelper, Search::Base, …).

But until we get rid of BaseHelper I think it's still a good improvement to generate code that is less noisy, @waiting-for-dev thoughts?

gsmendoza:
> The overridden helpers don't seem to be used elsewhere in solidus_{core, backend, api}, so why not including them from somewhere else that is not Spree::Base and remove the monkey patches altogether?

@waiting-for-dev I'm not sure what you mean about including the helpers from somewhere else that is not `Spree::BaseHelper`. Can you clarify?

In https://github.com/solidusio/solidus_starter_frontend/pull/243#discussion_r963391289, I tried the following but they weren't able to override Solidus Core’s `Spree::BaseHelper`:

* Copy the `Spree::BaseHelper` methods to ApplicationHelper.
* Copy the `Spree::BaseHelper` methods to a new local BaseHelper class and setting helper 'base' in StoreController.

I eventually found that I can override Solidus Core's `Spree::BaseHelper` by copying the whole `Spree::BaseHelper` file to `app/helpers`. However, this requires copying *all* of the `Spree::BaseHelper` methods, since if I remove any of them locally, the app wouldn't be able to find the corresponding method from Solidus Core. That wasn't ideal, so I ended up with the monkey patches solution from https://github.com/solidusio/solidus-demo/pull/103.

Reviewing https://edgeguides.solidus.io/customization/customizing-the-core/#using-overrides now, I'll do the following and test if this works:

1. Rename `monkey_patches` to `overrides`.
2. Remove the `monkey_patch` suffix from the `Spree::BaseHelper` override file.

waiting-for-dev:
Thanks for the context, @gsmendoza & @elia, I missed that comment.

I understand that the end goal is to get rid of `BaseHelper,` but that will take some time. What about if we rename the methods we use here? That way, we won't have any collision with that module. I know it's a bit of contortionism, but I think it's worth it to remove monkey patches from here. Thoughts?

elia:
@waiting-for-dev @gsmendoza given SSF is sourced from master I feel some urgency in merging this anyway and be iterative, providing a slightly better service to anyone installing from now until we have a better solution ready.

Thoughts on merging and then implementing the other improvements?

gsmendoza:
@elia @waiting-for-dev  I intend to remove the monkey patches and convert the helpers into view components, as initially suggested by @kennyadsl in https://github.com/solidusio/solidus_starter_frontend/pull/243#discussion_r963606659. Although view component might be a little bit heavy for some of the helpers, I think they're still much better than the monkey patch. There are only four helpers, so if I avoid refactoring for now, it should be a quick fix :)

waiting-for-dev:
Sure, we can merge it and iterate it. Yeah, if the components solution will be quick, it makes sense to go for it without renaming the methods.

template.rb
```diff
@@ -92,23 +92,33 @@ def copy_solidus_starter_frontend_files
   copy_file 'config/initializers/devise.rb', 'tmp/devise.rb'
   append_file 'config/initializers/devise.rb', File.read('tmp/devise.rb')
 
-  application <<~RUBY
-    if Rails.autoloaders.main
-      # Rails 7
+  if Rails.gem_version >= Gem::Version.new('7')
```
waiting-for-dev, line 95: Are we sure that will work for Rails 6 with Zeitwerk configured?

https://guides.rubyonrails.org/classic_to_zeitwerk_howto.html#why-switch-from-classic-to-zeitwerk-questionmark

gsmendoza:
Closing this PR in favor of https://github.com/solidusio/solidus_starter_frontend/pull/259.

PR #256, opened 2022-09-22
Fix the solidus core route being added twice

elia:
It looks like the route app-template command gets confused if there are multiple route blocks in config/routes.rb. 

## Description

See last commit on https://github.com/gsmendoza/solidus_store/tree/rails-7-0-4--solidus-3-2-2, removing a duplicate entry on a rails 6.1 app with solidus and SSF installed, that generated the double engine entry.

Using the `route` method to add the routes to the existing `draw` block should avoid the confusion of having two block to look at for rails 6.1.

## How Has This Been Tested?

Relying on the CI + local manual testing.

## Types of changes

## Checklist:


bin/sandbox
```diff
@@ -115,7 +115,7 @@ bundle exec rails generate solidus:install \
   --frontend=none \
   $@
 
-bundle exec app:template LOCATION="$PWD/../template.rb"
+bundle exec rails app:template LOCATION="$PWD/../template.rb"
```
kennyadsl, line 118: Shouldn't this go in the previous commit?

kennyadsl (commented):
The last commit still needs something, BTW, I like how we cleaned the installer, thanks Elia!

bin/sandbox
elia: Yes, sure, I was just checking if it was fixing the CI, but when I saw it was green I eagerly set the PR as ready for review 😅

kennyadsl (approved):
Thanks Elia!

PR #258, opened 2022-09-22
Remove associate_user

cpfergus1:
## Description

Removes `associate_user` from the orders_controller and the checkout_controller.

## Motivation and Context

See https://github.com/solidusio/solidus/pull/4201 for full context.
The method was used before, but now its responsibilities are fulfilled by the solidus_auth_devise which this repo depends on.

Some tests were modified to provide a user with an order being tested. Previously these tests assumed a user could have an order not associated with and navigate to a point in checkout for that order. In the checkout flow, the user will have an order associated with them after logging in which is fulfilled by a [hook on after_set_user in solidus_auth_devise](https://github.com/solidusio/solidus_auth_devise/blob/09841c164e1d0e6ab7afbb80494737fc692eb806/config/initializers/warden.rb#L4). The order association functionality is also[ tested within the solidus_auth_devise gem](https://github.com/solidusio/solidus_auth_devise/blob/master/spec/features/checkout_spec.rb#L73). 

## How Has This Been Tested?

The current testing suite covers changes - Manually tested checkout flow to ensure user is associated with the order

### DIFF
--- a/templates/app/controllers/checkout_controller.rb
+++ b/templates/app/controllers/checkout_controller.rb
@@ -14,7 +14,6 @@ class CheckoutController < StoreController
   before_action :ensure_sufficient_stock_lines
   before_action :ensure_valid_state
 
-  before_action :associate_user
   before_action :check_registration, except: [:registration, :update_registration]
   before_action :check_authorization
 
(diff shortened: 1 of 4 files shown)


elia (approved):
Good riddance! ✂️

waiting-for-dev (approved):
Lovely, @cpfergus1!

PR #260, opened 2022-09-26
Simplify newly-added components

gsmendoza:
## Background

I converted the `logo` and `seo_url` helpers in `Spree::BaseHelper` methods to components in https://github.com/solidusio/solidus_starter_frontend/pull/259. I originally created this PR to add unit tests for these components, but I realized that the components were simple enough to either inline or convert back into a helper.

* `LogoComponent` - I inlined this in the `top_bar` partial where it is used.
* `SeoUrlComponent` - I converted this back to a helper. Since it doesn't generate HTML markup, it's a bit overkill to render it as a component. However, I renamed the helper to `taxon_seo_url` to prevent it from clashing with the `seo_url` from `Spree::BaseHelper`. I also added a spec for this helper.

## Types of changes

~- [ ] Bug fix (non-breaking change which fixes an issue)~
~- [ ] New feature (non-breaking change which adds functionality)~
~- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)~

## Checklist:

~- [ ] My change requires a change to the documentation.~
~- [ ] I have updated the documentation accordingly.~


templates/spec/components/logo_component_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+      it 'renders the default logo image' do
+        expect(logo_image[:src]).to match(%r{/assets/logo/solidus-\S+\.svg})
+      end
+    end
+
+    context 'with an image_path' do
+      # For convenience, we're setting the argument to an existing image in the
+      # code base. Otherwise, we'd get an
+      # Sprockets::Rails::Helper::AssetNotFound for the image_path.
+      let(:image_path) { 'icon-delete.svg' }
```
gsmendoza, line 30: I couldn't find a way to stub the image asset. Any ideas?

waiting-for-dev: You can set `config.assets.unknown_asset_fallback` to `true` for this test. See https://github.com/rails/sprockets-rails/blob/master/README.md#initializer-options. The exception won't be raised in that case:

https://github.com/rails/sprockets-rails/blob/60f032e8a42e63a2179225392ecdb38f4ac06fa4/lib/sprockets/rails/helper.rb#L84

Otherwise, you could use a fixture image.

gsmendoza: Thank! @waiting-for-dev ! I'll give it a try :+1:

gsmendoza: @waiting-for-dev I ended up inlining the helper, so the spec is no longer necessary.

waiting-for-dev (approved):
Thanks, @gsmendoza!

PR #262, opened 2022-09-26
Alert user when TOS not checked

cpfergus1:
## Description

Adds a simple alert for the user if they have not checked the ToS box on the order confirmation page.
## Motivation and Context

When you place an order on the starter frontend, you need to check the "Agree to Terms of Service" checkbox before being able to proceed with the "Place Order" submit button.

If you just forget to check it, the "Place Order" button doesn't work but no error is rendered to the user.
We expect an informative message to be rendered to the user so they can see why they cannot submit the form.

## How Has This Been Tested?

Appended steps to a test to click on "place order" prior to accepting ToS and then accepting the alert. 

## Types of changes

## Checklist:

### DIFF
--- a/templates/app/assets/javascripts/checkout.js
+++ b/templates/app/assets/javascripts/checkout.js
@@ -18,17 +18,18 @@ window.addEventListener('DOMContentLoaded', () => {
   const termsCheckbox = document.getElementById('accept_terms_and_conditions');
 
   if (termsCheckbox) {
-    termsCheckbox.addEventListener('change', () => {
-      const submitButton = termsCheckbox.closest('form')
-        .querySelector('[type="submit"]');
-
+    const form = termsCheckbox.closest('form');
+    const submitButton = form.querySelector('[type="submit"]');
+    form.onsubmit = function () {
       if (termsCheckbox.checked) {
+        submitButton.innerHTML = 'Submitting...';
+        return true;
+      } else {
+        alert('Please review and accept the Terms of Service');
         submitButton.removeAttribute('disabled');
         submitButton.classList.remove('disabled');
-      } else {
-        submitButton.setAttribute('disabled', true);
-        submitButton.classList.add('disabled');
-      }
-    });
-  }
+        return false;
+      };
+    }; 
+  };
 });
--- a/templates/spec/system/checkout_spec.rb
+++ b/templates/spec/system/checkout_spec.rb
@@ -297,13 +297,16 @@ RSpec.describe 'Checkout', :js, type: :system, inaccessible: true do
       order.payments << create(:payment)
       visit checkout_state_path(:confirm)
 
+      # Test TOS not checked alert
+      accept_alert('Please review and accept the Terms of Service') { click_button "Place Order" }
+
       # prevent form submit to verify button is disabled
       page.execute_script("document.getElementById('checkout_form_confirm').onsubmit = function(){return false;}")
 
       check 'Agree to Terms of Service'
-      expect(page).not_to have_selector('button.button-primary[disabled]')
       click_button "Place Order"
-      expect(page).to have_selector('button.button-primary[disabled]')
+      button = find('button.button-primary')
+      expect(button).to be_disabled
     end
   end
 
--- a/templates/app/views/checkout/steps/_confirm_step.html.erb
+++ b/templates/app/views/checkout/steps/_confirm_step.html.erb
@@ -17,7 +17,6 @@
     I18n.t('spree.place_order'),
     class: 'button-primary',
     name: :commit,
-    disabled: true
   ) %>
 </div>
 
--- a/templates/spec/system/address_spec.rb
+++ b/templates/spec/system/address_spec.rb
@@ -62,10 +62,10 @@ RSpec.describe 'Address', type: :system, inaccessible: true do
           page.find(@state_name_css).set('Toscana')
 
           select france.name, from: 'Country'
-
-          expect(page).to have_no_css(@state_name_css)
-          expect(page).to have_no_css(@state_select_css)
         end
+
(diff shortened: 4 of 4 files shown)


templates/app/assets/javascripts/checkout.js
```diff
@@ -18,17 +18,19 @@ window.addEventListener('DOMContentLoaded', () => {
...
-      const submitButton = termsCheckbox.closest('form')
-        .querySelector('[type="submit"]');
-
-      if (termsCheckbox.checked) {
+    const form = termsCheckbox.closest('form')
+    const submitButton = form.querySelector('[type="submit"]');
+    form.onsubmit = function () {
+      if (!termsCheckbox.checked) {
+        termsCheckbox.scrollIntoView(); 
+        alert('Please review and accept the Terms of Service');
```
waiting-for-dev, line 26: We need to provide i18n here. I think we did something similar through `data-` attributes here:

https://github.com/solidusio/solidus_starter_frontend/blob/d3dd1e74e81cf8d6bd6dfe8cad68aa3e10c4851e/templates/app/assets/javascripts/product_selection_component.js#L17

templates/app/assets/javascripts/checkout.js
```diff
@@ -18,17 +18,19 @@ window.addEventListener('DOMContentLoaded', () => {
...
         submitButton.removeAttribute('disabled');
-        submitButton.classList.remove('disabled');
+        submitButton.classList.remove('disabled')
+        return false;
       } else {
-        submitButton.setAttribute('disabled', true);
-        submitButton.classList.add('disabled');
+        submitButton.innerHTML = 'Submitting...'
```
waiting-for-dev, line 31: Same here in regards to i18n.

waiting-for-dev (commented):
Thanks, @cpfergus1! That's nicely implemented. I just left some comments.

templates/app/assets/javascripts/checkout.js
```diff
@@ -18,17 +18,19 @@ window.addEventListener('DOMContentLoaded', () => {
...
-    termsCheckbox.addEventListener('change', () => {
-      const submitButton = termsCheckbox.closest('form')
-        .querySelector('[type="submit"]');
-
-      if (termsCheckbox.checked) {
+    const form = termsCheckbox.closest('form')
+    const submitButton = form.querySelector('[type="submit"]');
+    form.onsubmit = function () {
+      if (!termsCheckbox.checked) {
+        termsCheckbox.scrollIntoView();
```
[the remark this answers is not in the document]
cpfergus1: I thought about this as well, it makes sense to ignore the focus. I will remove it!

templates/app/assets/javascripts/checkout.js
```diff
@@ -18,17 +18,19 @@ window.addEventListener('DOMContentLoaded', () => {
...
-      if (termsCheckbox.checked) {
+    const form = termsCheckbox.closest('form')
+    const submitButton = form.querySelector('[type="submit"]');
+    form.onsubmit = function () {
+      if (!termsCheckbox.checked) {
+        termsCheckbox.scrollIntoView(); 
+        alert('Please review and accept the Terms of Service');
         submitButton.removeAttribute('disabled');
-        submitButton.classList.remove('disabled');
+        submitButton.classList.remove('disabled')
```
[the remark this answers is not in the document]
cpfergus1: The button is disabled by Solidus.disableSaveOnClick after clicking the "Place Order" button which triggers a form-submit event. This is to prevent a possible double-submit event. I had to clear it in case we do not want to submit the form so the user could click the button again. There does not appear to be any styling associated with disabled such as normally turning grey - I can look into this and make the stylistic changes but I think this also works well as is and maintains previous functionality.

templates/app/assets/javascripts/checkout.js
cpfergus1: 💯

templates/app/assets/javascripts/checkout.js
cpfergus1: 💯

templates/app/assets/javascripts/checkout.js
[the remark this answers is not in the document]
waiting-for-dev: Ok, I see. Feel free to ignore the comment in that case.

templates/app/assets/javascripts/checkout.js
waiting-for-dev: Forget about it, I see that everything is hardcoded to English, at this point.

templates/spec/system/checkout_spec.rb
```diff
@@ -297,13 +297,16 @@
...
+
       # prevent form submit to verify button is disabled
       page.execute_script("document.getElementById('checkout_form_confirm').onsubmit = function(){return false;}")
 
       check 'Agree to Terms of Service'
-      expect(page).not_to have_selector('button.button-primary[disabled]')
       click_button "Place Order"
-      expect(page).to have_selector('button.button-primary[disabled]')
+      button = find('button.button-primary')
+      expect(button.disabled?).to be_truthy
```
waiting-for-dev, line 309: Being fussy but if `button.disabled?` returns `true`, it'd communicate better to use the `be_true` matcher instead of `be_truthy` 🙂

waiting-for-dev (approved):
Thanks, @cpfergus1, minor non-blocking detail!

templates/spec/system/checkout_spec.rb
elia: I think it can even be `expect(button).to be_disabled`

cpfergus1: I was scared by Rubocop 👍

PR #265, opened 2022-10-03
Handle robots in here

elia:
Moving the changes to `robots.txt` from the solidus installer to here.

## Description

This was previously added by `solidus:install` but having to do with frontend routes it's better to handle it here.

The fact that `solidus:install` will keep appending these lines is not an issue as Thor's behavior won't append anything if it's already there.

## Motivation and Context

Those robots routes are coming from the frontend, if the frontend changes we should update them accordingly and it's better to do it from here.

## Types of changes

## Checklist:

### DIFF
--- a/template.rb
+++ b/template.rb
@@ -189,6 +189,17 @@ with_log['installing routes'] do
     get '/cart_link', to: 'store#cart_link', as: :cart_link
 
   RUBY
+
+  append_file "public/robots.txt", <<-ROBOTS.strip_heredoc
+    User-agent: *
+    Disallow: /checkout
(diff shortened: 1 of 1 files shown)


waiting-for-dev (approved):
Thanks!!

kennyadsl (approved):
Same as https://github.com/solidusio/solidus_frontend/pull/17#pullrequestreview-1128548817

elia:
@kennyadsl sent a reply there, meanwhile I'm merging this one, for solidus_frontend I don't have the power to do that

PR #266, opened 2022-10-04
Remove Empty Cart feature

gsmendoza:
Goal
----

As a maintainer of SolidusStarterFrontend
I want to remove the Empty Cart function
So that we won't have to maintain a secondary function that's not critical to the app
And we won't have to keep the custom route for that function

Background
----------

This is part of an epic to make the order and checkout routes more resourceful. 

This will be the planned outcome of that epic:

image

This is a 15 minute video explaining the epic: https://www.loom.com/share/c384938b2d224ebfaff03b0f0cf2d311. 

Note that this was recorded before we decided to remove the Empty Cart feature function altogether.

## Types of changes

## Checklist:

### DIFF
--- a/templates/app/assets/stylesheets/components/cart/_cart_secondary_actions.scss
+++ b/templates/app/assets/stylesheets/components/cart/_cart_secondary_actions.scss
@@ -19,7 +19,6 @@
     }
 
     &::after {
-      content: "or";
       margin-right: .5rem;
     }
 
--- a/template.rb
+++ b/template.rb
@@ -180,7 +180,6 @@ with_log['installing routes'] do
 
     get '/cart', to: 'orders#edit', as: :cart
     patch '/cart', to: 'orders#update', as: :update_cart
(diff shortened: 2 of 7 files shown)


templates/app/assets/stylesheets/components/cart/_cart_secondary_actions.scss
```diff
@@ -19,7 +19,6 @@
     }
 
     &::after {
-      content: "or";
```
kennyadsl, line 22: Maybe we can remove the whole `&::after`?

kennyadsl (approved):

waiting-for-dev:
Sorry, I think I missed that conversation. But why do we want to remove the empty cart feature? It looks like something that every store will need out there.

gsmendoza:
As discussed with @waiting-for-dev and @kennyadsl , we'll keep the `empty` action for now.

PR #269, opened 2022-10-07
Extract CartsController from OrdersController

gsmendoza:
## Goal

As a SolidusStarterFrontend user
I want to cart-related actions to be extracted from OrdersController to CartsController
So that it would be easier to follow the order checkout flow from cart to completion

## Background

This is part of an epic to refactor the order and checkout-related controller. Please see https://docs.google.com/spreadsheets/d/1mV2d4H4Ak9kEvc2bxKUpAv4hiz7FwTUE2t-SfOF7g8E/edit?usp=sharing for the list of changes.

## Types of changes

## Checklist:

- ~[ ] My code follows the code style of this project.~


templates/spec/requests/carts_spec.rb
```diff
@@ -11,29 +11,18 @@
...
 
       it 'renders the cart' do
-        get edit_order_path(order.number)
+        get edit_cart_path
 
         expect(flash[:error]).to be_nil
         expect(response).to be_ok
       end
-
-      context 'when another order number than the current_order' do
```
gsmendoza, line 20: I removed this spec because we don't use the orders/edit path to edit the cart.

templates/spec/requests/carts_spec.rb
```diff
@@ -55,32 +44,23 @@
...
         expect do
-          put order_path(order.number), params: { checkout: true }
+          patch cart_path, params: { checkout: true }
         end.to change { order.reload.state }.from('cart').to('address')
 
         expect(response).to redirect_to checkout_state_path('address')
       end
     end
-
-    context 'when order is not present' do
```
gsmendoza, line 76: I removed this spec because we don't use the `PUT order/` route to update the cart.

templates/app/controllers/orders_controller.rb
```diff
@@ -6,80 +6,19 @@ class OrdersController < StoreController
...
-      @order.empty!
-    end
-
-    redirect_to cart_path
-  end
+  private
 
   def accurate_title
-    if @order && @order.completed?
-      t('spree.order_number', number: @order.number)
```
gsmendoza, line 58: The new OrdersController only has the `show` action, so we can be sure that the order exists and has been completed.

templates/app/controllers/carts_controller.rb
```diff
@@ -0,0 +1,77 @@
...
+      @order.empty!
+    end
+
+    redirect_to edit_cart_path
+  end
+
+  private
+
+  def accurate_title
+    t('spree.shopping_cart')
```
gsmendoza, line 54: I removed the conditional when I copied `accurate_title` method from OrdersController to CartsController.

kennyadsl (approved):
Thanks @gsmendoza!

PR #277, opened 2022-11-04
Fix bundle env command being called on BundlerContext instead of Bundler

elia:
## Description

Since we're not testing the template when authentication is not installed this didn't came up in the CI.

This is a fixup of https://github.com/solidusio/solidus_starter_frontend/pull/276.

Thanks again to @stefano-sarioli for reporting this 🙌

## Motivation and Context

image

## Types of changes

## Checklist:

### DIFF
--- a/bin/sandbox
+++ b/bin/sandbox
@@ -93,13 +93,9 @@ bundle install --gemfile Gemfile
 
 bundle exec rake db:drop db:create
 
-bundle exec rails generate solidus:install \
-  --auto-accept \
-  --enforce_available_locales=true \
-  --frontend=none \
-  $@
-
-bundle exec rails app:template LOCATION="$PWD/../template.rb"
+export FRONTEND="$PWD/../template.rb"
+cp "$PWD/../solidus_3_2_patch.rb" config/initializers/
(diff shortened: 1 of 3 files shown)


kennyadsl (approved):
👍 for this temporary workaround. I hope we can enable the installer specs on Solidus Core v3.2 and remove the double auth_devise entry in the Gemfile. In the meantime, at least this allows having Solidus installed. Thanks for jumping on this @elia

waiting-for-dev:
For the record, that's what we discussed changing after this hotfix:

- Porting the `bin/sandbox` cleanup to `main`.
- Moving the patch on `Solidus::InstallGenerator::InstallFrontend` to Solidus v3.2.
- Revert https://github.com/solidusio/solidus/commit/3b7ec0afa13b131cd1e4b2dc6c7516fd36e1ed78. Leave SSF v3.2 without the responsibility to install solidus_auth_devise so we can remove the check on the Gemfile. Alternatively, update the installer on v3.2 to try to install solidus_auth_devise after the frontend.
- Release a new patch version for Solidus v3.2 to be consumed from here.

PR #279, opened 2022-11-11
Miscellaneous fixes

elia:
## Description

This is an extraction from https://github.com/solidusio/solidus_starter_frontend/pull/278 of things not strictly related to supporting solidus_paypal_commerce_platform and are of general utility.

## How Has This Been Tested?

CI

## Types of changes

## Checklist:


bin/rake
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env ruby
```
gsmendoza, line 1: Duplicate "standard" in "Add standard standard rake tasks and bin/setup" commit message.

template.rb
```diff
@@ -69,14 +69,23 @@
   gem 'truncate_html'
   gem 'view_component', '~> 2.46'
 
+  gem_group :test do
+    # Include the system tests gem generated by Rails on new installs with
+    # minitest, since many apps will be generated passing `--skip-test` we
+    # need to support both scenarios.
```
gsmendoza, line 75: Can you elaborate on this comment? It mentions several things (system tests gem, minitest, `skip-test`, "both scenarios"), but I don't understand how they relate with capybara and the other test gems below.

gsmendoza (changes requested):

Rakefile
```diff
@@ -0,0 +1,10 @@
+require 'rake/clean'
```
waiting-for-dev, line 1: For context, how did you generate these files?

bin/sandbox
```diff
@@ -24,12 +24,6 @@ case "${DB}" in
     ;;
 esac
 
-if [[ -n "${SOLIDUS_BRANCH}" ]]; then
```
waiting-for-dev, line 27: Could we add more information to the message of this commit? I think we're doing several things here:

- Installing after solidus
- Cleaning unneeded gems
- Acknowledging that solidus_starter_frontend's main branch only works for solidus master branch

waiting-for-dev: Ideally, we should split it in different commits, but a better explanation can be enough 🙂

templates/spec/support/solidus_starter_frontend/capybara.rb
```diff
@@ -2,40 +2,16 @@
...
-end
 
 RSpec.configure do |config|
   config.before(:each, type: :system) do
     driven_by((ENV['CAPYBARA_DRIVER'] || :rack_test).to_sym)
   end
 
   config.before(:each, type: :system, js: true) do
-    driven_by((ENV['CAPYBARA_JS_DRIVER'] || :apparition).to_sym)
+    driven_by((ENV['CAPYBARA_JS_DRIVER'] || :selenium_chrome_headless).to_sym)
```
waiting-for-dev, line 15: I'm ok removing apparition, but I'm afraid we're also losing docker support for tests here. Did you check if that's still working?

waiting-for-dev (changes requested):
Thanks, @elia. I'd left some considerations.

template.rb
elia: I'm replacing it with:

```
    # We need to add capybara along with a javascript driver to support the provided system specs,
    # `rails new` will add the following gems for system tests unless `--skip-test` is provided.
    # We want to stick with them but we can't be sure about how the app was generated, so we'll
    # add them only if they're not already in the Gemfile.
```

hope it makes more sense, but basically you'll get those three gems in your pristine `rails new` gemfile depending on which flags you provided. We also want them as they're the default for rails testing but we need to take into account that they might be already there.

Rakefile
elia: I created it manually, the `with_unbundled_env` part by trial and error.

In case you were interested specifically in `rake/clean`, this is standard Rake, and it provides the `clobber` task borrowing the term from `make`, see [Clobbering - Wikipedia](https://en.wikipedia.org/wiki/Clobbering)
https://en.wikipedia.org/wiki/Clobbering

template.rb
```diff
@@ -69,14 +69,24 @@
   gem 'truncate_html'
   gem 'view_component', '~> 2.46'
 
+  gem_group :test do
+    # We need to add capybara along with a javascript driver to support the provided system specs,
```
gsmendoza, line 73:
```suggestion
    # We need to add capybara along with a javascript driver to support the provided system specs.
```

Period instead of comma?

gsmendoza (changes requested):

template.rb
elia: Period is fine 👍 (pushed)

templates/spec/support/solidus_starter_frontend/capybara.rb
waiting-for-dev: Yeah, I think it's better to fix Docker on another PR. But let's rebase this one on top of the other (or wait until it's merged and rebase from main). I want to be sure that we're not breaking tests on Docker here, as I'm quite confident that removing the custom options will break it.

I've created the PR to fix Docker to speed up things and not block you. Please, could you check that https://github.com/solidusio/solidus_starter_frontend/pull/280 works for you? 🙏 I've cherry-picked your commit installing default-mysql-client, and I've also added libvips to the image. However, I can't reproduce your issue with nokogiri.

bin/sandbox
waiting-for-dev: Thanks!

template.rb
```diff
@@ -69,14 +69,24 @@
...
   gem 'view_component', '~> 2.46'
 
+  gem_group :test do
+    # We need to add capybara along with a javascript driver to support the provided system specs.
+    # `rails new` will add the following gems for system tests unless `--skip-test` is provided.
+    # We want to stick with them but we can't be sure about how the app was generated, so we'll
+    # add them only if they're not already in the Gemfile.
+    gem "capybara" unless Bundler.locked_gems.dependencies['capybara']
+    gem "selenium-webdriver" unless Bundler.locked_gems.dependencies['selenium-webdriver']
+    gem "webdrivers" unless Bundler.locked_gems.dependencies['webdrivers']
```
waiting-for-dev, line 79: I'd love to get rid of this hack, as that's something that end users will see. On top of that, I've experienced `Bundler.locked_gems` to be very slow sometimes.

What about modifying our spec helper with something like this for each case?

```ruby
begin
  require 'capybara'
rescue LoadError
  raise "SSF needs capybara to be available for.... Please add it to the the `:test` group in your Gemfile"
end
```

waiting-for-dev, line 79: I'd love to get rid of this hack, as that's something that end users will see. On top of that, I've experienced `Bundler.locked_gems` to be very slow sometimes.

What about modifying our spec helper with something like this for each case?

```ruby
begin
  require 'capybara'
rescue LoadError
  raise "SSF needs capybara to be available for.... Please add it to the the `:test` group in your Gemfile"
end
```

waiting-for-dev (changes requested):

templates/app/views/products/_product_info.html.erb
```diff
@@ -1,6 +1,6 @@
 <div class="product-page__description">
   <h2 class="product-description__title">
-    <%= t('description') %>
+    <%= t('spree.description') %>
```
waiting-for-dev, line 3: Awesome! I saw there were some warnings because of that while running tests ❤️

template.rb
elia: > I'd love to get rid of this hack, as that's something that end users will see.

This won't be seen by the user, as it's happening in the app-template, that `gem` DSL is not the same as the one we have in a Gemfile. The slowness probably happens when the lock is not there already and needs to be generated (just guessing).

The proposed solution could work but given we're not dirtying anything in the user Gemfile I would consider a better experience for everything to be working out of the box (at least for the most beaten paths).

waiting-for-dev: Oh, yeah, sorry, I got confused. Yeah, if that doesn't pollute the user's Gemfile, that's the best solution 👍 Sorry for the noise.

elia (commented):
@waiting-for-dev you think we can merge this and possibly fix the docker development env in a separate PR?

waiting-for-dev:
Hey @elia, I'm ok as long as we keep the custom options for the driver, although I'm ok getting rid of apparition. We need to fix Docker on ARM, but if we merge the as it is now we'll be introducing a regression for x86. At a later stage, we can also try to remove those custom options if we find a way. Does it make sense?

templates/spec/support/solidus_starter_frontend/capybara.rb
```diff
@@ -2,40 +2,31 @@
...
   end
 end
+
+# Taken from capybara's own chrome-headless driver with added
+# `opts.add_argument('--disable-featues=VizDisplayCompositor')`.
+Capybara.register_driver :solidus_starter_frontend_docker_chrome_headless do |app|
+  options_key = Capybara::Selenium::Driver::CAPS_VERSION.satisfied_by?(version) ? :capabilities : :options
```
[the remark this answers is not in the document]
elia: Sorry, I missed one line while copy/pasting from capybara, now should be fixed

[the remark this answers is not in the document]
elia: It seems to be working fine, I also added bin/rspec so it's easier to run failed specs from the rspec output

waiting-for-dev (approved):
We finally got rid of the apparition driver but kept the custom option so that Docker is not broken on x86. Fixing Docker for ARM is still pending.

PR #283, opened 2022-11-18
Fix uninitialized constant AppBase when monkey-patching InstallFrontend

gsmendoza:
The error is caused by
https://github.com/solidusio/solidus/commit/62ce6f2bc362fa7a2955fbbfb8677685f5ffef41. However, we're making the fix here in SolidusStarterFrontend because the issue only applies to this monkey-patch i.e. it doesn't break Solidus itself. Furthermore, we'll be removing this monkey-patch soon, so this issue will be removed with it.

See https://app.circleci.com/pipelines/github/solidusio/solidus_starter_frontend/1734/workflows/9c4b6afa-c53b-4f61-b449-f6aeb6c35965/jobs/5639 for an instance of this error.

## Types of changes

## Checklist:

### DIFF
--- a/solidus_3_2_patch.rb
+++ b/solidus_3_2_patch.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require 'rails/generators'
+require 'rails/generators/app_base'
 require 'generators/solidus/install/install_generator/install_frontend'
 
 module SSFSolidus32Patch


elia (approved):
ha! I was just looking in that CI failure, great to see it's already fixed 🙌

kennyadsl (approved):

PR #289, opened 2022-12-21
Allow creating the layout with importmap support

kennyadsl:
## Description

When creating the main host app with `rails new`, we are used to do that with the `--skip-javascript` flag. This flag was essentially skilling all the webpack stuff that we were not supporting.

Now that Rails defaults to the old asset pipeline in combination with importmaps, there's really no reason to keep building applications with that flag, and we should start supporting (and using?) this tools.

When the application is made with JS, our layout will have the directive that allows to use the importmaps, as any other Rails application.

## How Has This Been Tested?

To test this, I created a new sandbox application, and verified that the layout contained the new `importmap` directive. 

## Types of changes

## Checklist:


templates/app/views/layouts/storefront.html.erb.tt
```diff
@@ -0,0 +1,31 @@
+<!doctype html>
+<html lang="<%%= I18n.locale %>">
+  <head>
+    <meta charset="utf-8">
+    <title><%%= title %></title>
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <%%= meta_data_tags %>
```
waiting-for-dev, line 7: Are those `<%%=` typos from `<%=`?

kennyadsl: No, that's a way to let the `.tt` file render the erb tags without processing them. Close to the `importmap` directive, you can see an occurrence of a tag that is instead interpreted and won't be shown in the view on the host application.

waiting-for-dev: Oh, I see. Thanks.

waiting-for-dev (approved):

templates/app/views/layouts/storefront.html.erb.tt
```diff
@@ -0,0 +1,30 @@
...
+    <%%= stylesheet_link_tag 'solidus_starter_frontend', media: 'screen' %>
+
+<% if Rails.application.respond_to? :importmap -%>
+    <%%= javascript_importmap_tags %>
+<% end -%>
+
+    <%%= favicon_link_tag 'favicon.ico' %>
+    <%%= yield :head %>
+  </head>
+  <body x class="<%%= body_class %>" id="<%= @body_id || 'default' %>">
```
gsmendoza, line 21: @kennyadsl I notice that the x attribute was added in this commit. Is it a typo?

kennyadsl: Ops, yes, it was a test I was doing to understand if my changes were immediately reflected when rebuilding the sandbox. BTW, they weren't: I needed to create a commit to see those changes applied.

Thanks a lot for spotting this.

PR #294, opened 2023-01-04
Remove Order Transitions Spec

cpfergus1:
## Description
This Spec Modifies the Spree::Order class by adding an after_transition callback which bleeds over into other specs utilizing Spree::Order model. This causes flakey specs depending on if they run after this specific test is run. This spec appears to originate from Spree as seen here:
https://github.com/spree/spree/issues/2004#issuecomment-12703434

## Motivation and Context
We should not be modifying Classes like this in specs because of the potential to cause unintended failures.

## Types of changes


kennyadsl:
Also, this spec has almost nothing to do with the frontend and should not be a request spec. 👍

PR #296, opened 2023-01-13
Prevent New Line Before Colon After Payment Name

cpfergus1:
## Description
There appears to be an oversight where adding the colon on the outside of the "p" tag caused a new line break which puts the colon on the line directly below the payment method name - ex:
```
payment method name

:
amount
```

This commit adds the colon with in the p tag so that they will display on the same line. Another way this could be approached is to use span and then style it, but I felt this was the most practical and simple method for correcting this issue.

## Motivation and Context
Payment Method Display looks bad with the all those line breaks

## How Has This Been Tested?
Visually

|Before|After|
|-----|-----|
|image|image|

## Types of changes

## Checklist:

### DIFF
--- a/templates/app/views/orders/_payment_info.html.erb
+++ b/templates/app/views/orders/_payment_info.html.erb
@@ -22,7 +22,7 @@
         <li><%= source.name %></li>
       </ul>
     <% elsif source.is_a?(Spree::StoreCredit) %>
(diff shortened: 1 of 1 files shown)


elia (approved):

elia:
(force-pushed in an attempt to trigger the CI)

PR #297, opened 2023-01-13
Respond with 400 with invalid search params

kennyadsl:
## Motivation and Context

It is better not to respond with 500, both to avoid polluting error monitoring services, and because it's formally more correct. In fact, malformed user inputs should not raise a 500 error.

This will fail until we merge https://github.com/solidusio/solidus/pull/4844. As soon as it's merged we have the new exception available and we can use it. 

## How Has This Been Tested?

Locally and with specs.

## Types of changes

## Checklist:


elia (approved):
✔️

PR #298, opened 2023-01-20
Rename route from /cart/edit to /cart

elia:
## Description

Since GET /edit is just a specialized "show" action in REST and our cart page has both roles, we should go toward the better looking URL, which is incidentally also familiar to everyone.

## Types of changes

## Checklist:


waiting-for-dev (approved):
Makes sense.

Non-blocking nit: shouldn't we squash both commits? The first can't work without the second.

gsmendoza (approved):

elia:
@waiting-for-dev I kept them separate because the last one is very noisy and would have drawn the change in the first one among all those renames, so I optimized for reviewing.

Happy to "Squash and merge" or if you feel strongly about that I'll push a squashed commit 👍

waiting-for-dev:
> @waiting-for-dev I kept them separate because the last one is very noisy and would have drawn the change in the first one among all those renames, so I optimized for reviewing.
> 
> Happy to "Squash and merge" or if you feel strongly about that I'll push a squashed commit 👍

It's ok, @elia, the policy at Solidus is not strict here.

PR #300, opened 2023-02-02
Autocomplete main search with products and taxons

kennyadsl:
Adds some very basic autocomplete functionalities to our main search input. It returns results for what has been typed so far for products and taxons, up to 5 for each type.

For products, it also allows users to see all results for the given keyword, falling back to the standard search. 

Once the suggestions box appears, some very basic UX allows to navigate results:

- Key down: go to the next result.
- Key up: go to the previous results.
- Enter: open the currently selected result.
- Esc: close the suggestions box.
- Click anywhere else on the page: close the suggestions box.

Screenshot 2023-02-02 at 14 52 31

Thanks @elia for guidance and help during its development.

## Description

This feature uses stimulus to submit the form on every key pressed in the search input (if > 2 chars) and add some UX to navigate the results.

Turbo stream will take care of updating the results section with the suggestions retrieved from the new specific controller action introduced.

## Motivation and Context

We want to provide a basic implementation for stores that want to add autocomplete capabilities to their store. This solution is basic but highly customizable based on each store's specific needs.

## How Has This Been Tested?

Tested manually and via specs (see PR changes). 

## Types of changes

## Checklist:

### DIFF
--- a/templates/app/controllers/autocomplete_results_controller.rb
+++ b/templates/app/controllers/autocomplete_results_controller.rb
@@ -1,0 +1,40 @@
+# frozen_string_literal: true
+
+class AutocompleteResultsController < StoreController
+  def index
+    respond_to do |format|
+      format.html { redirect_to products_path(keywords: params[:keywords]) }
+      format.turbo_stream { load_results }
+    end
+  end
+
+  private
+
+  def load_results
+    @results ||= begin
+      results = {}
+      results[:products] = autocomplete_products
+      results[:taxons] = autocomplete_taxons
+      results
+    end
+  end
+
+  def autocomplete_products
+    if params[:keywords].present?
+      searcher = build_searcher(params.merge(per_page: 5))
+      searcher.retrieve_products
+    else
+      Spree::Product.none
+    end
+  end
+
+  def autocomplete_taxons
+    if params[:keywords].present?
+      Spree::Taxon
+        .where(Spree::Taxon.arel_table[:name].matches("%#{params[:keywords]}%"))
+        .limit(5)
+    else
+      Spree::Taxon.none
+    end
+  end
+end
--- a/templates/app/javascript/controllers/search_controller.js
+++ b/templates/app/javascript/controllers/search_controller.js
@@ -1,0 +1,58 @@
+import { Controller } from "@hotwired/stimulus"
+
+export default class extends Controller {
+  static targets = [ "form", "keywords", "results", "result"]
+  static classes = [ "current" ]
+
+  // This is needed to restore the current result index when the
+  // results are updated after a search.
+  resultsTargetConnected() {
+    this.currentResultIndex = 0
+    this.render()
+  }
+
+  fetchResults() {
+    if(this.keywords.length < 2) {
+      this.reset()
+      return
+    }
+
+    clearTimeout(this.timeout)
+    this.timeout = setTimeout(() => this.formTarget.requestSubmit(), 500)
+  }
+
+  nextResult() {
+    if(this.currentResultIndex < this.resultTargets.length - 1) {
+      this.currentResultIndex++
+      this.render()
+    }
+  }
+
+  previousResult() {
+    if(this.currentResultIndex > 0) {
+      this.currentResultIndex--
+      this.render()
+    }
+  }
+
+  openResult() {
+    this.resultTargets[this.currentResultIndex].firstElementChild.click()
+  }
+
+  reset() {
+    this.currentResultIndex = 0
+    if (this.hasResultsTarget) {
+      this.resultsTarget.innerHTML = ""
+    }
+  }
+
+  render() {
+    this.resultTargets.forEach((element, index) => {
+      element.classList.toggle(this.currentClass, index == this.currentResultIndex)
+    })
+  }
+
+  get keywords() {
+    return this.keywordsTarget.value
+  }
+}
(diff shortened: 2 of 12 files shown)


templates/app/assets/stylesheets/components/search/_search_bar.scss
```diff
@@ -34,3 +34,49 @@
     padding: 0 2rem 0 1rem;
   }
 }
+
```
waiting-for-dev, line 37: Isn't this CSS duplicated with what we have in the `_autocomplete_results.scss` file?

templates/spec/system/search_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+    fill_in 'keywords', with: 'shirt'
+    click_button 'Search'
+
+    expect(page.all('ul.products-grid li').size).to eq(1)
+  end
+
+  context 'with JS enabled', :js do
+    it 'shows autocomplete suggestions' do
+      fill_in 'keywords', with: 'ruby'
+      expect(page.all('ul.autocomplete-results__list li.autocomplete-results__item:not(.autocomplete-results__item--empty)').size).to eq(8)
```
waiting-for-dev, line 22: I'm a bit concerned about how these tests rely so heavily on classes used by CSS. Thoughts about leveraging `data-` attributes?

waiting-for-dev (commented):
Thanks, @kennyadsl. I left two comments. It looks like CI is failing for older branches. Is that related to #302?

templates/app/assets/stylesheets/components/search/_search_bar.scss
kennyadsl: lol, yes, I started there, then moved to a different file and forgot to remove it, thanks!!!

templates/spec/system/search_spec.rb
kennyadsl: Always unsure about how to handle this problem, but I like the idea, especially if we are testing some JS-driven behavior that will change when we change those `data-` attributes. Do you know if there's any best practice for what to use in specs?

waiting-for-dev: > Do you know if there's any best practice for what to use in specs?

Not sure. If possible, we should rely on behavior instead of concrete UI implementation. If not possible, I think we should try to leverage HTML semantics and, if that's neither possible, I sometimes have used `data-test` attribute names. I think there's no perfect solution, though 😢

kennyadsl: Done. I just left a check on the "current" CSS class on results, which is what the JS does to mark the currently selected item.

templates/app/javascript/controllers/search_controller.js
```diff
@@ -0,0 +1,60 @@
...
+  previousResult() {
+    if(this.currentResultIndex > 0) {
+      this.currentResultIndex--
+      this.render()
+    }
+  }
+
+  openResult() {
+    this.resultTargets[this.currentResultIndex].firstElementChild.click()
+  }
```
elia, line 59: nit: I think these three are not "private", would also be fine to remove the comment, give how little it means in this context

templates/app/views/autocomplete_results/_autocomplete_results.html.erb
```diff
@@ -0,0 +1,36 @@
...
+      </li>
+    <% end %>
+
+    <li class="autocomplete-results__item"
+        data-search-target="result">
+      <%= link_to "See all results", products_path(keywords: params[:keywords]) %>
+    </li>
+  <% else %>
+    <li class="autocomplete-results__item autocomplete-results__item--empty">No results</li>
+  <% end %>
```
elia, line 20: Looks like the only difference between this block and the following one is the translation key and the results url/name. Thoughts on refactoring that to a loop or partial?

elia (approved):
🙌

templates/app/views/autocomplete_results/_autocomplete_results.html.erb
kennyadsl: One difference is that the products list has the `See all results` item while the taxons list doesn't (taxons don't have an index page, filterable by a search term). 

That was enough for me to determine that an abstraction here wasn't the right thing, but let me know if you think we should move something in a partial anyway.

templates/app/javascript/controllers/search_controller.js
kennyadsl: Thanks!

elia (approved):

templates/app/controllers/autocomplete_results_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+    else
+      Spree::Product.none
+    end
+  end
+
+  def autocomplete_taxons
+    if params[:keywords].present?
+      Spree::Taxon
+        .where(Spree::Taxon.arel_table[:name].matches("%#{params[:keywords]}%"))
+        .limit(5)
```
gsmendoza, line 35: Non-blocking: the search implementation in `autocomplete_taxons` is on a lower-level compared to `autocomplete_products`. Perhaps `build_searcher` can be updated to handle this query as well?

gsmendoza (approved):

waiting-for-dev (approved):

templates/app/controllers/autocomplete_results_controller.rb
kennyadsl: @gsmendoza Yes, that's a good idea, even if a bit out of scope for this PR. At the moment the searcher class used by `build_searcher` assumes you are [searching for a Product](https://github.com/solidusio/solidus/blob/d7ef0f23164a1e5821e4da1cd316d39c6ea96f79/core/lib/spree/core/search/base.rb#L36) and changing this would require a big refactor.

PR #302, opened 2023-02-02
CI Strategy Redesign

kennyadsl:
## Description

Revisit the CI strategy to reflect the environments where this starter kit is installed. This PR adds some documentation about the compatibility with Solidus and the versioning testing strategy we want to use. 

Scheduled Workflows have been disabled because it's not possible to schedule jobs on several branches with different parameters. For example, we can't use a different Rails version for different branches. My plan to solve this problem is by using [CircleCI Scheduled Pipelines](https://circleci.com/blog/using-scheduled-pipelines/), which allows scheduling jobs on different branches with specific parameters. The downside is that we cannot use a configuration file to set the schedule parameters. 

## Motivation and Context

Right now, the jobs we are running in the CI do not reflect the state of the project. In fact, we are branching the main branch at each release and making the Solidus installation target the corresponding version of Solidus Starter Frontend (see [v3.3](https://github.com/solidusio/solidus/pull/4898) and [v3.2](https://github.com/solidusio/solidus/blob/v3.2/core/lib/generators/solidus/install/install_generator/install_frontend.rb#L46)). 

## Next Steps

- [ ] Do the same changes to v3.3
- [x] Configure daily Scheduled Pipeline for branch `main`
- [x] Configure daily Scheduled Pipeline for the latest released branch (currently `v3.3`)
- [ ] Add a new step to the Solidus release guide: at every new Solidus release, we need to change the Scheduled Pipeline to point to the newly released branch. (text ready: `- Change [Latest Version Nightly build Scheduled Pipeline's trigger](https://app.circleci.com/settings/project/github/solidusio/solidus_starter_frontend/triggers) to point to the newly created branch.`)

## How Has This Been Tested?

CI is passing! Scheduled Pipeline approach [has been tested](https://github.com/solidusio/solidus_starter_frontend/pull/302#issuecomment-1415933287). 

## Types of changes

## Checklist:


README.md
```diff
@@ -78,6 +92,37 @@ For information about contributing to this project please refer to this
...
+
+The following parameters are considered in the testing strategy
+in the CI for this project:
+### Database
+
+We are testing this starter kit against PostgreSQL and MySQL.
+### Ruby Version
+
+We are testing this starter kit against the latest Ruby version
+and the last minor of the previous major. Eg: 3.2 and 2.7.
```
waiting-for-dev, line 105: Shouldn't we test against ruby versions that are still maintained, regardless of the major/minor number? For instance, once v3.3 is out, and v2.7 is EOL, we should test for v3.0, v3.1 & v3.2. That'd also be aligned with our new policy for main Solidus.

README.md
```diff
@@ -78,6 +92,37 @@ For information about contributing to this project please refer to this
 * Releasing new versions
 * Docker development
 
+## CI/Testing Strategy
+
+The following parameters are considered in the testing strategy
+in the CI for this project:
+### Database
+
+We are testing this starter kit against PostgreSQL and MySQL.
```
waiting-for-dev, line 101: Thoughts on adding SQLite to test development? We're doing that on the main Solidus and at least some extensions.

README.md
```diff
@@ -78,6 +92,37 @@ For information about contributing to this project please refer to this
...
+### Ruby Version
+
+We are testing this starter kit against the latest Ruby version
+and the last minor of the previous major. Eg: 3.2 and 2.7.
+
+### Rails version
+
+We are testing this starter kit against the last Rails version
+at the time the corresponding Solidus version has been released.
```
waiting-for-dev, line 110: Once Rails v7.1 is out and we're still on v3.3, that would translate into not covering the latest Rails, which will also be the most common use case. What do you think about testing against the latest Rails supported by any Solidus version?

README.md
```diff
@@ -78,6 +92,37 @@ For information about contributing to this project please refer to this
...
+Eg. Solidus v3.3 tests against Rails 7.0
+
+### Solidus Version
+
+- `main` branch will test installing itself over Solidus master only.
+- each `vX.X` branch will only test installing itself over the
+  corresponding Solidus version.
```
waiting-for-dev, line 117: I guess we only need to test the current `vX.X` branch, for older ones it doesn't make sense.

README.md
kennyadsl: Yes, you are right!

README.md
kennyadsl: But Solidus `v3.3` won't be supporting Rails `v7.1` and we would need to wait for Solidus `v3.4` to be released. That why I said `the last Rails version at the time the corresponding Solidus version has been released`. Does it sound correct? Or are you proposing to change the supported version exactly when Rails releases new versions?

waiting-for-dev: You're right, I was confused.

kennyadsl: Still, we need to be more clear about the Rails versions supported in the installation guides: when Rails 7.1 is out, it's legit to expect people to use it with a fresh Solidus installation.

kennyadsl:
Trying out the CircleCI scheduled pipelines, I configured a trigger for the latest versions (now targeting `v3.3`):

Screenshot 2023-02-03 at 15 15 20@2x

Which seems to work well:

Screenshot 2023-02-03 at 15 17 40@2x

README.md
kennyadsl: But... When Rails 7.1 is out, Solidus `v3.3` [won't be compatible with that version of Rails](https://github.com/solidusio/solidus/blob/v3.3/core/solidus_core.gemspec#L30). So formally, in terms of dependencies, we are safe.

.circleci/config.yml
```diff
@@ -65,7 +69,7 @@ commands:
           environment:
             RAILS_VERSION: <<parameters.rails_version>>
             SOLIDUS_BRANCH: <<parameters.branch>>
-            TEST_RESULTS_PATH: ../test-results/gems-v3-ruby-v2-7-solidus-<<parameters.branch>>/results.xml
+            TEST_RESULTS_PATH: ../test-results/ruby-<<parameters.rails_version>>-rails-<<parameters.rails_version>>-solidus-<<parameters.branch>>-<<parameters.database>>-results.xml
```
elia: Long overdue! 👏

.circleci/config.yml
```diff
@@ -109,24 +109,24 @@ workflows:
     jobs:
       - run-specs:
           context: slack-secrets
-          name: run-specs-with-postgres
+          name: run-specs-with-postgres-ruby-3-2
```
elia, line 112: Thoughts on removing defaults and always listing the database and ruby_version?

elia (commented):
Looks great!

Left a couple of comments

.circleci/config.yml
```diff
@@ -107,37 +107,27 @@ jobs:
...
       - run-specs:
           context: slack-secrets
-          name: run-specs-with-solidus-3-2-rails-7-ruby-3-1-postgres
+          name: run-specs-with-postgres
 
       - run-specs:
           context: slack-secrets
-          name: run-specs-but-with-solidus-latest
-          solidus_branch: 'master'
+          name: run-specs-with-mysql
```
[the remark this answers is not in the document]
kennyadsl: But we don't know how many versions of ruby will be supported together, now they are three. Is that a constant? Maybe yes, because they are released once per year.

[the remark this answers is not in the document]
elia: Yeah, surely not blocking, it's something I've done in the past and worked well, your call 👍

.circleci/config.yml
kennyadsl: Good idea. Going to update this.

waiting-for-dev (approved):
Thanks 👍

PR #314, opened 2023-02-10
Abort installation when Rails < 7

kennyadsl:
## Description

We now have features that are only compatible with the last version of Rails and we are assuming the project will be only installed on the latest.

This change let the installer exit if a previous version is detected, printing a message that asks to retry with latest Rails.

## Motivation and Context

We don't want Solidus Starter Frontend to be installed with old versions of Rails, otherwise new features won't work.

## How Has This Been Tested?

I changed the sandbox script to use an old version of Rails:

```
rails _6.1.7.2_ new sandbox \
  --skip-git \
  --database=${DB:-sqlite3} \
  --skip-rc
```

The result is Rails and Solidus installed, but the Solidus Starter Frontend template script aborted with the corresponding message. 

## Types of changes

## Checklist:

### DIFF
--- a/template.rb
+++ b/template.rb
@@ -6,12 +6,12 @@ with_log = ->(message, &block) {
 }
 
 with_log['checking versions'] do
-  if Rails.gem_version < Gem::Version.new('6.1')
-    say_status :unsupported, shell.set_color(
-      "You are installing solidus_starter_frontend on an outdated Rails version.\n" \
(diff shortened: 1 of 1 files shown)


gsmendoza (approved):

waiting-for-dev (approved):

elia (approved):
👍 

This is yet another situation in which having dependency requirements would have helped 🤔

PR #318, opened 2023-02-15
Fix clicking on autocomplete results

### LINKED ISSUE #313: Autocomplete search result does not respond to mouse click
## Solidus Version:

3.4.0.dev

## To Reproduce

1. It seems we currently can't run this repo's sandbox app because of https://github.com/solidusio/solidus/pull/4933. So, to run the sandbox app, you might need to checkout out the `gsmendoza/4932-fix-sandbox-error branch` from https://github.com/nebulab/solidus and run `bin/rails server` from there. 
2. Go to https://localhost:3000 and enter a search term in the search field.
3. Try clicking on a search result.

See this video:

https://user-images.githubusercontent.com/61476/217981582-fb687e86-0792-485f-8fa1-141130e766b1.mp4

## Current behavior

Nothing happens. However, I can select a search result by hitting `enter` with the keyboard.

## Expected behavior

I should be able to able to select a search result by clicking on the result.

## Desktop (please complete the following information):
 - OS:  Pop! OS 22.04 LTS
 - Browser: Google Chrome 
 - Version: 110.0.5481.77 (Official Build) (64-bit)

## Additional context

The autocomplete feature was introduced in https://github.com/solidusio/solidus_starter_frontend/pull/300.

kennyadsl:
## Description

Fixes #313

Follow-up for #300 

## Motivation and Context

By clicking on one of the results item, the `focusout` event is triggered and it acts before the `click` event is triggered, removing the list.

To solve this, we listen to all click events and do nothing if the click is internal to the autcomplete, otherwise it resets it.

## How Has This Been Tested?

By automated tests and in the sandbox locally

## Types of changes

## Checklist:

### DIFF
--- a/templates/spec/system/search_spec.rb
+++ b/templates/spec/system/search_spec.rb
@@ -39,6 +39,15 @@ RSpec.describe 'searching products', type: :system do
       expect(page.all('[data-search-target="result"]')[0][:class]).to include('autocomplete-results__item--current')
     end
 
+    it 'allows mouse click on results' do
+      fill_in 'keywords', with: 'ruby'
+      # wait for the autocomplete to show up
+      expect(page.all('[data-search-target="result"]')[0][:class]).to include('autocomplete-results__item--current')
+      find_all('[data-search-target="result"]')[0].click
+
+      expect(page).to have_current_path('/products/ruby-on-rails-ringer-t-shirt')
+    end
+
     it 'clicks on a suggestion pressing enter' do
       fill_in 'keywords', with: 'ruby'
       # wait for the autocomplete to show up
--- a/templates/app/javascript/controllers/search_controller.js
+++ b/templates/app/javascript/controllers/search_controller.js
@@ -39,6 +39,12 @@ export default class extends Controller {
     this.resultTargets[this.currentResultIndex].firstElementChild.click()
   }
 
+  focusOut(event) {
+    if (!this.formTarget.contains(event.target)) {
+      this.reset()
+    }
+  }
+
   reset() {
     this.currentResultIndex = 0
     if (this.hasResultsTarget) {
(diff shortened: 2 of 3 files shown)


elia (approved):

gsmendoza (approved):

gsmendoza:
Thanks @kennyadsl . I've confirmed that it's working :)

waiting-for-dev (approved):

templates/spec/system/search_spec.rb
```diff
@@ -39,6 +39,15 @@
       expect(page.all('[data-search-target="result"]')[0][:class]).to include('autocomplete-results__item--current')
     end
 
+    it 'allows mouse click on results' do
+      fill_in 'keywords', with: 'ruby'
+      # wait for the autocomplete to show up
+      expect(page.all('[data-search-target="result"]')[0][:class]).to include('autocomplete-results__item--current')
```
waiting-for-dev, line 45: Are you sure that the `include` RSpec matcher will [wait](https://github.com/teamcapybara/capybara#asynchronous-javascript-ajax-and-friends) until the item is present?

waiting-for-dev (commented):
Thanks, @kennyadsl. I left a question as I'm not sure the new test could become a flaky candidate.

templates/spec/system/search_spec.rb
kennyadsl: I think it's more `page.all` that is waiting here, from the [`#all` doc](https://rubydoc.info/github/jnicklas/capybara/master/Capybara%2FNode%2FFinders:all):

> By default Capybara's waiting behavior will wait up to [default_max_wait_time](https://rubydoc.info/github/jnicklas/capybara/master/Capybara#configure-class_method) seconds for matching elements to be available and then return an empty result if none are available.

So, you are right that this is not ideal, meaning that we would probably have the same effect without the `include('autocomplete-results__item--current')` part. 

Because this is used in multiple tests over this file, if that's ok for you I'd merge this and shoot another PR that cleans this up in the whole file.

waiting-for-dev: Makes sense 👍

PR #321, opened 2023-02-16
Payment step cleanup

elia:
## Summary

Removes the two tier radio buttons to select between wallet payment sources and filling in from scratch.

It's predicated under the following assumptions:

- The previous system was confusing, with its interacting radio buttons which were presented on both sides without much consistency.
- The list gives precedence to wallet payments as they require less clicks and were already the default choice if present.
- If no wallet payments are available the first payment method is selected, according to the order set in admin.
- Since to the customer they're all payment methods they're treated as a single radio list.
- Both types of payments are now visually consistent, better communicating all the options to the customer.
- The payment details are wrapped in a fieldset that gets disabled whenever a payment method is unselected, this also works as a hook for the payment provider specific implementation (e.g. solidus_stripe) to react properly when selected/unselected.

This layout and decisions are also informed by multiple real world stores that got their payment step redesigned.

- depends on #301 

### Before
<details>
<summary>
Show original behavior
</summary>

https://user-images.githubusercontent.com/1051/219420532-c91ffe24-ea19-488b-84f4-8ba45145195e.mov

image
</details>

### After

image

https://user-images.githubusercontent.com/1051/219440642-112350bc-1893-459c-b4fd-c8f24aba4e44.mov

## Checklist

- [x] I have kept my commits small and atomic.


kennyadsl:
@elia still not reviewing but... a "before" screenshot would be helpful to quickly navigate what's changing, especially when the description of the PR is empty. 😬

elia:
@kennyadsl yes 💯 agree, will probably do a short video demonstrating how broken it is

templates/app/views/checkouts/payment/_check.html.erb
```diff
@@ -0,0 +1 @@
+<%= payment_method.class.model_name.human %>
```
kennyadsl, line 1: I think Payment Method has a description field, we could use that.

templates/app/javascript/controllers/checkout_payment_controller.js
```diff
@@ -0,0 +1,30 @@
+import { Controller } from "@hotwired/stimulus"
```
kennyadsl, line 1: ❤️

kennyadsl (approved):
Left one quick improvement idea, but overall: YES!

PR #328, opened 2023-03-14
Turn the product-card partial into a component

elia:
## Summary

- Only render the currency microdata when the price is present
- Keep the cache block in the products grid
- Add specs for rendering with/without a taxon

## Checklist


templates/app/views/products/_products-grid.html.erb
```diff
@@ -10,6 +10,13 @@
 
 <ul class="<%= class_names %>">
   <% products.each do |product| %>
-    <%= render 'products/product-card', product: product %>
+    <%= cache [I18n.locale, current_pricing_options, @taxon, product] do %>
+      <%= render ProductCardComponent.new(
+        product,
+        taxon: @taxon,
+        class_name: local_assigns.fetch(:classes, []).join(" "),
```
kennyadsl, line 17: Does this pass the current partial classes to the inner component?

kennyadsl (commented):
Nice, left a question about a line that I don't understand.

templates/app/views/products/_products-grid.html.erb
elia: I think so, this is just a way to keep what was there before, specifically on this line:

https://github.com/solidusio/solidus_starter_frontend/blob/7cf7e76d54256cf90e5ab56d1c7805ca84d0a1db/templates/app/views/products/_product-card.html.erb#L8

Although maybe it doesn't make sense in the starter frontend, as nowhere this partial was being called passing an extra `classes` local variable.

kennyadsl: Correct me if I'm wrong, but I think it's different. 

In the partial, it's using the `local_assigns`, which uses any local variable passed when initializing the partial. But we weren't passing any `classes` local variable when rendering the partial (`<%= render 'products/product-card', product: product %>`). 

With the current code instead, we are passing the local classes to the inner component.

elia: Oh good call! Removing it ✂️

templates/spec/components/product_card_component_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+
+    aggregate_failures do
+      expect(rendered_component.css("li .product-card_header a").to_html).to include("The best product")
+      expect(rendered_component.css("li .product-card_header a").first[:href]).to eq('/products/123')
+      expect(rendered_component.css("li .product-card_image a").first[:href]).to eq('/products/123')
+    end
+  end
+
+  context 'with a taxon' do
+    it "renders the product with its main image and price" do
```
kennyadsl, line 21: Maybe we can mention in the scenario description that the link will differ in this case?

kennyadsl (approved):
Nice, left a possible improvement, but I'm fine with that.

templates/spec/components/product_card_component_spec.rb
elia: @kennyadsl done, although we should probably remove passing a taxon to the product card component, really looks like a legacy of the older frontend

kennyadsl: Agree, I don't even know in which cases it could be useful.

elia:
@kennyadsl added a commit removing support for `@taxon` ✂️

kennyadsl (approved):

PR #329, opened 2023-03-14
Track spec coverage with codecov

elia:
## Summary

- Coverage is reported for the `sandbox/` dir (I wasn't able to trick SimpleCov into believing those paths were in `templates/`)
- Enabled only for 3.2 which is able able track ERB files

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -34,6 +34,8 @@ commands:
         type: string
       database:
         type: string
+      coverage:
+        type: boolean
 
     steps:
       - run:
@@ -46,7 +48,9 @@ commands:
           when: always
       - run:
           name: 'Solidus <<parameters.solidus_branch>>:  Install sandbox with generated starter frontend'
-          command: bin/sandbox --seed=false --sample=false
+          command: |
+            <<#parameters.coverage>>export COVERAGE=true<</parameters.coverage>>
+            bin/sandbox --seed=false --sample=false
           environment:
             RAILS_ENV: development # if run in test mode will attempt to eager-load and break the sandbox
             RAILS_VERSION: <<parameters.rails_version>>
@@ -57,6 +61,7 @@ commands:
           command: |
             cd sandbox
             bundle add rspec_junit_formatter --group test
+            <<#parameters.coverage>>export COVERAGE=true<</parameters.coverage>>
             bundle exec rspec --format progress --format RspecJunitFormatter --out ../test-results/results.xml
           environment:
             RAILS_VERSION: <<parameters.rails_version>>
@@ -79,6 +84,8 @@ jobs:
           rails_version: <<parameters.rails_version>>
           ruby_version: <<parameters.ruby_version>>
           database: <<parameters.database>>
+          coverage: <<parameters.coverage>>
+
       - solidusio_extensions/store-test-results
     parameters:
       solidus_branch:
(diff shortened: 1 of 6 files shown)


.circleci/config.yml
```diff
@@ -57,6 +61,7 @@ commands:
           command: |
             cd sandbox
             bundle add rspec_junit_formatter --group test
+            <<#parameters.coverage>>export COVERAGE=true<</parameters.coverage>>
```
waiting-for-dev, line 64: Is that `<<#` an `if` statement on CircleCI :heart:? Do you have a reference for it? I looked for it, but I didn't find it.

.circleci/config.yml
```diff
@@ -46,7 +48,9 @@ commands:
           when: always
       - run:
           name: 'Solidus <<parameters.solidus_branch>>:  Install sandbox with generated starter frontend'
-          command: bin/sandbox --seed=false --sample=false
+          command: |
+            <<#parameters.coverage>>export COVERAGE=true<</parameters.coverage>>
```
waiting-for-dev, line 52: Do we need coverage when creating the sandbox application? :thinking:

bin/sandbox
```diff
@@ -39,6 +39,11 @@ fi
 
 cd ./sandbox
 
+# Add support for tracking coverage
+unbundled bundle add simplecov -v 0.22 --require=false --skip-install
+unbundled bundle add codecov --require=false
+echo "require_relative '../../coverage.rb' if ENV['COVERAGE']" >> config/boot.rb
+
```
waiting-for-dev, line 46: I'm confused. Shouldn't it go into the dummy-app? :thinking:

waiting-for-dev (approved):
Thanks, @elia!! :test_tube:  I left some comments & questions.

.circleci/config.yml
elia: Yes, it's an `if`, and I have to search for it deep in the circleci docs every time, there's an example here, although somewhere there must be a place with a syntax reference: https://circleci.com/docs/reusing-config/#commands-usage-examples

.circleci/config.yml
elia: In theory we would like to track usage of `template.rb` too, but I think it's still not fully working and will need additional passes once we get `codecov` going with a baseline on the main branch

bin/sandbox
elia: The starter was sort of the pioneer in this approach and does everything with the sandbox, eventually we'll want to normalize things across repositories and extract some tooling I guess, but for now that's the setup 😒

.circleci/config.yml
waiting-for-dev: Oh, yeah, that's really hidden :slightly_smiling_face: Thanks for sharing!

bin/sandbox
waiting-for-dev: Oh, I see. Thanks for clarifying.

.circleci/config.yml
waiting-for-dev: Not sure if getting that. When we run `bin/sandbox`, we're not running any spec file, right? How can `COVERAGE` make a difference there? :thinking:

elia: Coverage tracking works independently of having specs being run, just tracks which lines are executed during the life of a program. That means that if `template.rb` is executed after `Coverage.start` we'll get a list of lines for it in `Coverage.result`.

Since our coverage setup is "activated" by the `COVERAGE` env var we what it set when we install solidus and end up running the `template.rb` file. 

I hope it's clearer, but happy to jump on a call and go into the details.

PR #348, opened 2023-06-29
Handle checkout process for orders fully covered by store credits

### LINKED ISSUE #345: Preventing available payment method rendering for orders fully covered by Store Credits
## Short-Term Solution

**Desired Behavior**

We want to prevent the rendering of available payment methods if the order will be fully covered by store credits. The primary issue is that having a $0 checkout due to full coverage by store credits can interfere with the regular checkout process. (See https://github.com/solidusio/solidus_stripe/issues/307).
Our plan is to display a clear message indicating that the order is fully covered by store credits and remove the need for additional payment method selection.

**Current Behavior**

Currently, if an order is fully covered by store credits, the system still permits selection of a payment method. This often results in payments being created in the checkout state, which can get invalidated when store credits are added.

Screenshot 2023-05-25 at 17 29 31

**Additional notes**

Please note that there are edge cases where this solution may not cover the payment invalidation problem.
For example, if a user selects a payment method and then moves from the payment step to the confirm step, the issue may still arise if the user goes back to the payment step and selects and fills another payment method.

For more details about the long-term solution, refer to https://github.com/solidusio/solidus_starter_frontend/issues/346.

For more information about these proposed solutions, please refer to https://github.com/solidusio/solidus/issues/5109#issuecomment-1576600115.
This issue contains additional discussion and context about the identified problems and the suggested solutions to improve the store credit flow in Solidus.

rainerdema:
## Summary

Fixes #345

This PR adds logic to handle the checkout process for orders that are fully covered by store credits. 
In the checkout process, we skip the payment selection stage if the order can be entirely covered by store credits.
A message is displayed to the user indicating that no additional payment method is required.

These changes ensure a smoother checkout experience for users who have sufficient store credits to cover their entire order.

Screenshot 2023-06-29 at 14 52 19

Screenshot 2023-06-29 at 15 28 05

## Checklist

### DIFF
--- a/templates/app/views/checkouts/steps/_payment_step.html.erb
+++ b/templates/app/views/checkouts/steps/_payment_step.html.erb
@@ -3,60 +3,63 @@
     <legend>
       <%= t('spree.payment_information') %>
     </legend>
+    <% if @order.covered_by_store_credit? %>
+      <p>Your order is fully covered by store credits, no additional payment method is required.</p>
+    <% else %>
+      <ul class="payment-step__selector">
+        <% @wallet_payment_sources&.each do |wallet_payment_source| %>
+          <% payment_method = wallet_payment_source.payment_source.payment_method %>
+          <% default = wallet_payment_source == @default_wallet_payment_source %>
+          <% fieldset_name = "wallet-source-#{wallet_payment_source.id}" %>
 
-    <ul class="payment-step__selector">
-      <% @wallet_payment_sources&.each do |wallet_payment_source| %>
-        <% payment_method = wallet_payment_source.payment_source.payment_method %>
-        <% default = wallet_payment_source == @default_wallet_payment_source %>
-        <% fieldset_name = "wallet-source-#{wallet_payment_source.id}" %>
+          <li>
+            <label>
+              <%= radio_button_tag(
+                "order[wallet_payment_source_id]",
+                wallet_payment_source.id,
+                default,
+                'data-action': 'checkout-payment#paymentSelected',
+                'data-checkout-payment-target': "paymentRadio",
+                'data-fieldset-name': fieldset_name,
+                id: ","
+              ) %>
+              <%= t(payment_method.name, scope: 'spree.payment_methods', default: payment_method.name) %>
+            </label>
+            <fieldset name="<%= fieldset_name %>" class="payment-step__details">
+              <%= render(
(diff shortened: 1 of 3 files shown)


templates/app/views/checkouts/steps/_payment_step.html.erb
```diff
@@ -3,60 +3,63 @@
     <legend>
       <%= t('spree.payment_information') %>
     </legend>
+    <% if @order.covered_by_store_credit? %>
+      <p>Your order is fully covered by store credits, no additional payment method is required.</p>
```
rainerdema, line 7: This message needs to be associated with a translation key.

rainerdema: https://github.com/solidusio/solidus/pull/5208

elia: as pointed out by @kennyadsl in solidusio/solidus#5208 we can start storing frontend related keys in the starter itself, with time we should end up with a tighter, more coherent list of translations for the core instead of the giant ball of unused ones we currently have.

elia (changes requested):

elia:
@rainerdema I asked for moving the translation to here, but also wanted to point out that it's great to see this fixed 🙌

templates/app/views/checkouts/steps/_payment_step.html.erb
kennyadsl: True. Still, I'd like to keep that change separate from this PR. This change will end up in the starter fronted 4.2 so I think the current approach of adding the translation key to the core is safe because Solidus Starter Frontend 4.2 will only be installed on Solidus 4.2. 

I think we can go as Rainer suggested and open an issue for moving the translation keys from the core over here (I expect the usual backward compatibility pain 😅 ).

elia: @kennyadsl I think there are ways in which it can be less painful, e.g.:

- move all the deprecated translations in core to a separate locale files (e.g. `solidus/core/config/locales/en.deprecated.yml`)
- copy the file to the host application as part of the solidus installation/upgrade process
- remove the deprecated file in the next major

elia (approved):
After an internal discussion the hardcoded string is fine, and the starter specific locale file will be handled in a different issue/PR

👍👍👍

kennyadsl (approved):
👍

PR #350, opened 2023-07-06
Cleanup test files copied over the host application

kennyadsl:
## Summary

There were multiple inconsistencies and outdated code. Please read commit by commit for more details. 

## Checklist


templates/spec/solidus_starter_frontend_spec_helper.rb
```diff
@@ -85,6 +85,7 @@ module SolidusStarterFrontend; end
 
     example.run
 
+    Rails.cache.clear
```
elia, line 88: This seems to be doing a slightly different thing than the original, probably not an issue, unless it will show up only for some execution sequences.

For 100% compatibility I suggest moving it to a `config.before { Rails.cache.clear }`.

elia (approved):
All great ✨ 

I left a couple of optional comment, the one about clearing the cache deserving particular attention.

templates/spec/solidus_starter_frontend_spec_helper.rb
waiting-for-dev: Or we can move it above the `example.run` line, as we're within an `around(:each)` block.

waiting-for-dev (approved):
:clap:

templates/spec/solidus_starter_frontend_spec_helper.rb
elia: @waiting-for-dev that would probably work as well, although this `around` block is targeting examples with metadata containing `caching: true`.

kennyadsl: > that would probably work as well, although this around block is targeting examples with metadata containing caching: true.

Right, but if the cache is disabled for other specs, does it matter? I think that line where it was before is a bit like shooting a mosquito with a bazooka and we can do better.

elia (approved):

kennyadsl:
@elia @waiting-for-dev everything addressed, waiting for the CI before merging, thanks!

PR #368, opened 2024-06-19
Make changing the language work

mamhoff:
There were a few things that needed to be done here:
- Add stimulus attributes to the form in the footer
- Fix the stimulus controller to actually run the submit function
- Fix the cart link controller to consider a cart link stale if the locale changes
- Fix the link to cart component to translate the string "Cart"


kennyadsl (approved):
Thanks Martin!

PR #386, opened 2024-11-29
Replace dependency on `canonical-rails` with tag helper

forkata:
## Summary

This change removes the dependency on the `canonical-rails` gem by

- Adding additional canonical tag test coverage
- Implementing a simple canonical tag helper
- Removing dependency on `canonical-rails`

Considering we are not using some of the features provided by the gem, we decied
to reimplmented the feautres we are using in a simple tag helper. This will
handle the way the starter frontend currently uses the canonical tag helper,
but for storefronts that have more complex requirements, they can add the gem
as a dependency and continue using that.

## Upgrade Notes

Usage before upgrading

```
canonical_tag("example.com")
```

after upgrading

```
simple_canonical_tag(host: "example.com")
```

## Checklist

- [x] I have kept my commits small and atomic.


jarednorman (approved):

kennyadsl (approved):
Nice thanks! A rebase is needed here. ❤️

fthobe:
@forkata could you rebase so it can be merged <3

forkata:
@kennyadsl done. Looks like one test failed, but I don't think it's related to the changes here.

fthobe:
The test on search fails frequently probably to lack in wait time adjustment for the interface to come up. It's just flaky.

benjaminwil:
Thanks @fthobe. Could you open this as a new issue?

This pull request simply re-implemented what was always done by `canonical-rails` by default, without the dependencies. So I don't think a revert will satisfy anyone.

I agree that `simple_canonical_tag` does not solve the problem of duplicated content and does not holistically deal with Rails routing people to duplicated routes or not-actually-canonical pages. It is only meant to spit out a canonical tag. A developer can change the method in the application if the default settings are not acceptable. Or, as the documentation we added suggests: they can install `canonical-rails` and configure it properly and delete `simple_canonical_tag` entirely.

But it seems like you're suggesting that `canonical-rails` itself is not good enough.

fthobe:
@benjaminwil absolutely. 

I open up a new issue.

forkata:
@fthobe As @benjaminwil mentioned no behaviour was changed in this PR, this refactor just removed the dependency on `canonical-rails` and replaced it with a helper method that will be installed by the generator into the host application, which would allow you to customize that in any way you wanted. You can look at the specs in the PR if you want to understand what behaviour the new approach provides, which was meant to demonstrate what was used previously by `canonical-rails` and re-implemented only those parts of the library. Users who have further customized this behaviour or already depend on `canonical-rails` should be including that in their application's Gemfile and continue doing what they were doing previous to this change.

fthobe:
@forkata @benjaminwil 

There's some stuff happening to prepare a permanent solution for canonical.

#413 is reducing the product availability to 1 url (which as a byproduct currently bugfixes the canonical) in a not very elegant way by redirecting the product id and historic slugs of a product to the products current slug

[Solidus PR #6100](https://github.com/solidusio/solidus/pull/6100) brings redirects of historic links also to taxon permalinks (which in reality are also slugs, they just have the wrong name)

So step by step we are getting there. The point with the ash remains though:

Let me throw you this ball: let's assume for a moment that I am right about the backslash stuff (looks great but plays out terrible just like Anna Kurnikova and tennis) what can we do to solve this three issues: 

1. When people migrate to solidus they will either have all urls with a trailing backslash, none having a trailing backslash or any (more or less ietf compliant) mixture of having and not having backslashes (we as also candlescience are trailing backslash only people for example as it reduces SEO mistakes in data entry)

2. Usually you want to keep or better improve SEO (which usually means touching as few urls as any possible avoiding also changing paths to include /products/ or /t/). 

3. Not destroy performance of url requests. 

What can we do to solve for all three migration cases (with / without / mixed) the trailing "/" issue in one go? I'd be totally willing to hear to either collaborate or align with a feasible solution and put @shahmayur001 on it or maybe you come up with a plan that you want to implement by yourself!

PR #402, opened 2024-12-20
$ hard coded in collections with call to action

magpieuk:
Simple update to remove hard coded $ and instead use the display_price method to maintain shop currency.

### DIFF
--- a/templates/app/views/home/_collections_with_call_to_action.html.erb
+++ b/templates/app/views/home/_collections_with_call_to_action.html.erb
@@ -22,7 +22,7 @@
               <%= product.name %>
             </span>
             <span class="text-body-sm md:text-body font-sans-md price selling">
-              $<%= product.price %>
+              <%= product.display_price %>
             </span>
           </div>
         </div>


fthobe:
@magpieuk Hey, did you find more of them lyring around in the code?

magpieuk:
Not so far.

fthobe:
test postgres is failing related to your test: 

https://app.circleci.com/pipelines/github/solidusio/solidus_starter_frontend/3487/workflows/1a1c2a2e-2af1-4678-87e6-2afe0744ee97/jobs/14645

Finished in 5 minutes 54 seconds (files took 12.92 seconds to load)
324 examples, 3 failures, 14 pending

Failed examples:

```
rspec ./spec/system/products_spec.rb:221 # Visiting Products does not allow to put a product without a current price in the cart
rspec ./spec/system/products_spec.rb:47 # Visiting Products meta tags and title returns the correct title when displaying a single product
rspec ./spec/system/products_spec.rb:68 # Visiting Products meta tags and title doesn't use meta_title as heading on page
```

You have to update the circleci files, please also test manually the impact of your change in an additional currency and publish a screenshot as there probably also need to be adjusted some classes regarding the change in currency styling.

fthobe:
Finished in 5 minutes 54 seconds (files took 12.92 seconds to load)
324 examples, 3 failures, 14 pending

Failed examples:

rspec ./spec/system/products_spec.rb:221 # Visiting Products does not allow to put a product without a current price in the cart
rspec ./spec/system/products_spec.rb:47 # Visiting Products meta tags and title returns the correct title when displaying a single product
rspec ./spec/system/products_spec.rb:68 # Visiting Products meta tags and title doesn't use meta_title as heading on page

fthobe:
@magpieuk Hey, if you feel that fixing the testing is out of your reach, an alternative would be to grab the current store currency. 

An example can be found here: [Product Header Currency](https://github.com/solidusio/solidus_starter_frontend/blob/9f6d3974f9f088269a653296e5cd125363308063/templates/app/views/products/_product_header.html.erb#L20)

fthobe:
@magpieuk Do you want to finish this up? It was a good catch. Otherwise we fix it along with some other things we have on the plate.

magpieuk:
Sorry I will take a look at the tests. Looks like the currency is hard coded in them as well

fthobe:
That's not great. @kennyadsl I think we need your 5 cents (pun intended) here!

kennyadsl (approved):

kennyadsl:
I don't think spec failures are related, seems more like flaky specs. Re-running them to see if they pass.

fthobe:
@kennyadsl I do not believe it's smart to connect currency and price on the frontend, I think the $ should be fixed instead giving more flexibility.

kennyadsl (approved):

kennyadsl:
@fthobe that's what this PR does, I was just saying there's nothing hardocoded in the specs, we should be good to go. BTW, those homepage blocks are just a reference implementation, I don't think anyone is using them without changing their code. Still, great to have the template fixed.

@magpieuk thanks for the contribution!

fthobe:
@magpieuk hey, I would like to weight in on this topic with a little context: 

These are a collection of services that require you to reproduce the price with a separate ISO currency: 
- google analytics
- google shopping 
- google structured data 
- Facebook marketplace

I really appreciate what you do here but I believe that keeping the currency separated but not hardcoded is the way to go to later on keep them in two components. I would really appreciate if we could do that here to not create the need for additional variables but manage with two instead of three where ever the price and currency need to be displayed in context of a product / cart / order confirmation. 

Can we manage to use a way similar to this: 

```<h1 class="font-serif-md text-h4 lg:text-h3 mt-12 mb-3 lg:mt-3" itemprop="name">
    <%= product.name %>
  </h1>

  <% if product.price_for_options(current_pricing_options)&.money and !product.price.nil? %>
    <h2 class="font-sans-md text-body-md" id="product-price">
      <%= content_tag(
        :span,
        display_price(product),
        itemprop: 'price',
        content: product.price_for_options(current_pricing_options)&.money.to_d,
        data: { js: 'price' }
      ) %>

      <%= content_tag(
        :span,
        nil,
        itemprop: 'priceCurrency',
        content: current_pricing_options.currency
      ) %>
    </h2>
  <% end %>```

fthobe:
@magpieuk @kennyadsl we will never the less need two components that can be called individually for price and currency. I would also default in starter to use ISO three letter currencies everywhere.

tvdeyen (approved):

PR #403, opened 2024-12-20
#401Images vanishing

### LINKED ISSUE #401: Storefront images disapear
Images on store front disappear after a while. 

**Solidus Version**:4.4.2

When deployed to production environment with caching enabled, images disappear after a while.

src: image_tag image.url(size), default_options.merge(options)

image_tag url_for(image.attachment.variant(size)), default_options.merge(options)

magpieuk:
## Summary

  Fixes #401 

Images disappear when running in production mode with caching enabled.

### DIFF
--- a/templates/app/components/image_component.rb
+++ b/templates/app/components/image_component.rb
@@ -12,7 +12,7 @@ class ImageComponent < ViewComponent::Base
 
   def call
     if image
-      image_tag image.url(size), default_options.merge(options)
+      image_tag url_for(image.attachment.variant(size)), default_options.merge(options)
     else
       content_tag :div, nil, class: ['image-placeholder', size].join(' ')
     end


jarednorman (changes requested):
This solution doesn't work for other active storage backends. It only fixes it for local, which you probably shouldn't be using.

jarednorman (approved):
I was mistaken. According to [the docs](https://guides.rubyonrails.org/active_storage_overview.html#serving-files) this solution is supposed to generate a permanent URL. I'm curious as to how that would work backends like S3, but this seems like right approach.

fthobe:
@jarednorman we are installing the first prod similar env this week and can try it before anybody merges this.

fthobe:
@magpieuk is this a permanent setup for you or will you add any external media hosting in production like S3?

fthobe:
@magpieuk i will test it today. Can you tell me more about the environment that produced the problems?

boomer196:
This won't work if you have Solidus `image_attachment_module` configured to use `Spree::Image::PaperclipAttachment`.  There is no `variant` method defined on that class.

fthobe:
I am not sure if paperclip is still supported. I think that's something to be asked on slack and to be aligned on docs afterwards.

jarednorman:
Paperclip is still supported.

